PR #238, opened 2020-08-13
Remove Unsupported Value Known Issue

savannahostrowski:


jakebailey:
Only feedback would be to leave the `# Known Issues` section there, but empty, so we can add to it later if needed.

savannahostrowski:
You got it!

PR #280, opened 2020-08-25
Add note about VSCode variables not working

meichthys:
Currently VSCode variables do not work in Pylance. This should be documented.


meichthys:
Will do

PR #357, opened 2020-09-10
Update CHANGELOG.md

bschnurr:
updating changelog for 2020.9.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2020.9.2 (10 September 2020)
 
 Notable changes:
--   Bug Fix: Addressed memory and cpu issues a number of users had with indexer but setting it off by default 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,36 @@
 # Changelog
 
+## 2020.9.1 (10 September 2020)
+
+Notable changes:
+-   Bug Fix: Addressed memory and cpu issues a number of users had with indexer but setting it off by default
```
huguesv, line 6: but -> by

bschnurr: so previous line: Pylance will now index libraries and unopened files at startup to provide auto-import completions even for variables that have not been fully analyzed. This index is also used to improve the performance of the workspace symbols search.


- Addressing memory and cpu issues a number of users had by no longer indexing libraries and unopened files at startup.  This will revert auto-import completions and workspace symbols performance to previous levels.

huguesv (approved):

PR #985, opened 2021-02-24
Update GIF

savannahostrowski:
Copied the way the core extension does their GIFs.


README.md
```diff
@@ -12,7 +12,7 @@ This repository is for providing feedback and documentation on the Pylance langu
 
 # Features
 
-![ screencast ](screencap.gif)
+<img src=https://raw.githubusercontent.com/microsoft/pylance-release/images/all-features.gif>
```
jakebailey, line 15: I'd leave this as relative, then in our VSIX we can make it absolute to try and fix the bug.

savannahostrowski: Done deal

README.md
```diff
@@ -12,7 +12,7 @@ This repository is for providing feedback and documentation on the Pylance langu
 
 # Features
 
-![ screencast ](screencap.gif)
+<img src=images/all-features.gif>
```
jakebailey, line 15: Why the HTML tag and the the markdown tag as was used previously? The image tag works, but curious about the change.

jakebailey (approved):

PR #2479, opened 2022-03-16
Add workflow for daily report

gramster:


gramster:
Going to run these in a separate repo to not clutter things up here.

PR #2998, opened 2022-07-05
Create wikibot.py

sher2odbek:
none

### DIFF
--- a/wikibot.py
+++ b/wikibot.py
@@ -1,0 +1,47 @@
+
+import logging
+import wikipedia
+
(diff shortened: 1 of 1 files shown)


debonte:
@sher2odbek, perhaps you intended to send this PR to a different repo? If not, can you explain what you are trying to accomplish?

wikibot.py
```diff
@@ -0,0 +1,47 @@
+
+import logging
+import wikipedia
```
softorangetech11, line 3: this looks like wikipedia, this may be for wikipedia i guess?

PR #3299, opened 2022-09-07
CHANGELOG.md and README.md

debonte:


README.md
```diff
@@ -55,6 +55,18 @@ Pylance provides users with the ability to customize their Python language suppo
         -   `workspace`
         -   `openFilesOnly` (default)
 
+-   `python.analysis.include`
+    -   Paths of directories or files that should be included. If no paths are specified, Pylance defaults to the directory that contains workspace root. Paths may contain wildcard characters `**` (a directory or multiple levels of directories), `*` (a sequence of zero or more characters), or `?` (a single character).
+    -   Default value: empty array
+
+-   `python.analysis.exclude`
+    -   Paths of directories or files that should not be included. These override the include directories, allowing specific subdirectories to be excluded. Note that files in the exclude paths may still be included in the analysis if they are referenced (imported) by source files that are not excluded. Paths may contain wildcard characters `**` (a directory or multiple levels of directories), `*` (a sequence of zero or more characters), or `?` (a single character). If no exclude paths are specified, Pylance automatically excludes the following: `**/node_modules`, `**/__pycache__`, `.git` and any virtual environment directories.
+    -   Default value: empty array
```
debonte, line 64: The default value of exclude is an empty array, but we interpret that as `**/node_modules`, etc. Are you guys happy with the way it's written? Should I move the node_modules, etc list down to the default value line? Not sure what will be most clear to readers.

bschnurr: I think it makes sense as written

bschnurr (approved):

PR #3331, opened 2022-09-13
[WIP] adding indexing section

heejaechang:


README.md
```diff
@@ -92,6 +92,27 @@ Pylance provides users with the ability to customize their Python language suppo
...
+-   `python.analysis.autoImportCompletions`
+
+    -   Used to control the offering of auto-imports in completions.
+    -   Accepted values:
+        -   `true` (default)
+        -   `false`
+
+-   `python.analysis.packageIndexDepths`
+
+    -   Used to control how much installed packages indexer will index.
```
rchiodo, line 104:
```suggestion
    -   Used to control how many installed packages indexer will index.
```

README.md
```diff
@@ -92,6 +92,27 @@ Pylance provides users with the ability to customize their Python language suppo
         - `true`
         - `false`
 
+-   `python.analysis.autoImportCompletions`
```
bschnurr, line 95: `python.analysis.autoImportsEnabled` ?

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
 
     -   Convert import format according to `python.analysis.importFormat`.
 
+# Auto/Add import and indexing
+
+Pylance has a feature called ```auto import``` which will show symbols from installed packages and user files in completion. When committed, insert ```import``` statement for you. It also has a similar feature called ```add import``` that will show symbols from installed packages and user files for an unknown identifier. When selected, insert ```import``` statement for you.
```
rchiodo, line 242:
```suggestion
Pylance has a feature called ```auto import``` which shows symbols from installed packages and user files during completion. When committed, pylance will automatically generate an ```import``` statement. It also has something called ```add import``` which shows symbols from installed packages and user files for an unknown identifier. When selected, this will also insert an ```import``` statement.
```

README.md
heejaechang: option name is pre-existing one. I don't think we want to change it. it is not something I just added.

heejaechang: I just moved the option up in the README file so that auto import related options are listed together.

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+Since these features can be costly to scan symbol information and show them in completion and code actions, by default, we have minimal functionality enabled and provide multiple settings to control its behavior.
```
rchiodo, line 244:
```suggestion
Since these features can be costly, we have enabled the minimum set and provide multiple settings to control the behavior.
```

heejaechang:
thank you @rchiodo

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+When ```python.analysis.indexing``` is off, we will only show symbols already used by opened files transitively. If ```indexing``` is on, we will spend some time indexing installed packages and user files when pylance is started, but after that, we will provide symbols that you haven't used yet in completion and code action.
+
+But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which package they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
```
rchiodo, line 248:
```suggestion
But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which packages they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
```

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+For user files, we will index up to 2000 files. One difference between user files and installed packages is that they won't appear in completion unless it is already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion.
```
rchiodo, line 250:
```suggestion
For user files, we index up to 2000 files. One difference between user files and installed packages is that they won't appear in completion unless they are already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion. 
```

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which package they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
+
+For user files, we will index up to 2000 files. One difference between user files and installed packages is that they won't appear in completion unless it is already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion. 
+
+We choose to show them only in code action since it requires explicit user action ```Ctrl+.``` rather than automatically triggered like completion.
```
rchiodo, line 252: What is the 'them' you're referring to in this paragraph? Completions?

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+For people who don't want ```auto import``` suggestion ever show up in completion since they feel it bloats the completion, we have ```python.analysis.autoImportCompletions``` to disable it. And solely rely on code action.
```
rchiodo, line 254:
```suggestion
For people who don't want ```auto import``` suggestions to ever show up in completions since they feel it bloats the completions, you can use ```python.analysis.autoImportCompletions``` to disable it. Code actions can still be used with this setting disabled..
```

README.md
heejaechang: by them, I meant symbols from user files.

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+Since these features can be costly, we have enabled the minimum set and provide multiple settings to control the behavior.
+
+When ```python.analysis.indexing``` is off, we will only show symbols already used by opened files transitively. If ```indexing``` is on, we will spend some time indexing installed packages and user files when pylance is started, but after that, we will provide symbols that you haven't used yet in completion and code action.
+
+But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which packages they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
+
+For user files, we index up to 2000 files. One difference between user files and installed packages is that they won't appear in completion unless they are already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion.
```
rchiodo, line 250:
```suggestion
For user files, we will index up to 2000 files. One difference between user files and installed packages is that user files won't appear in completion unless they are already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion. 
```

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which packages they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
+
+For user files, we index up to 2000 files. One difference between user files and installed packages is that they won't appear in completion unless they are already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion. 
+
+We choose to show them only in code action since it requires explicit user action ```Ctrl+.``` rather than automatically triggered like completion.
```
rchiodo, line 252:
```suggestion
We show user file completions only in the code action since it requires an explicit user action ```Ctrl+.``` rather than being automatically triggered like completion.
```

README.md
heejaechang: I think it should be symbols from user files?

README.md
heejaechang: here as well. it is not user files, but all symbols from user files. here symbol include the file itself as a module. but also, all class and function defined in that module.

README.md
rchiodo: Yeah that sounds good

README.md
rchiodo: That sounds good

rchiodo: The 'them' was ambiguous to me so just wanted to fix that part.

README.md
```diff
@@ -92,6 +92,27 @@ Pylance provides users with the ability to customize their Python language suppo
...
+-   `python.analysis.autoImportCompletions`
+
+    -   Used to control the offering of auto-imports in completions.
+    -   Accepted values:
+        -   `true` (default)
+        -   `false`
+
+-   `python.analysis.packageIndexDepths`
+
+    -   Used to control how many installed packages indexer will index.
```
luabud, line 104: maybe "Used to control how many subfolders under installed packages should be indexed." or something like that? Just to reflect what depth would mean.

heejaechang: so, depth is not subfolder but nested modules. simply put, number of ```dot``` + 1

ex) ```package\__init``` is basically ```import package``` so it is depth 1
```package\lib.py``` is ```import package.lib``` so it is depth 2 even if it is on same folder?

not sure how to express it...

Comment (debonte) [reply README.md:104]: "Used to override the depth of indexing on a per package basis. Depth is measured in blah"?

File: README.md [hunk 12]
```diff
@@ -92,6 +92,27 @@ Pylance provides users with the ability to customize their Python language suppo
...
+    -   Default value: empty array
+    -   Accepted values: 
+        -   Tuple of [```package name```, ```depth to scan```, ```include all symbols```]
```
debonte, line 107: What happens when `include all symbols` is set to false?

heejaechang: we only include symbols in ```__all__```, if include all symbol is ```true```, we index every symbol decl in the file.

README.md
heejaechang: nice. it would be even better if you can provide the ```blah``` part as well :)

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+When ```python.analysis.indexing``` is off, we will only show symbols already used by opened files transitively. If ```indexing``` is on, we will spend some time indexing installed packages and user files when pylance is started, but after that, we will provide symbols that you haven't used yet in completion and code action.
+
+But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which packages they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
+
+For user files, we will index up to 2000 files. One difference between user files and installed packages is that symbols from user files won't appear in completion unless they are already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion.
```
debonte, line 250: It's not clear that the last sentence is explaining what would happen if we didn't limit user symbols to those already in use. Maybe something like this?
```suggestion
We will index up to 2000 user files, and unlike installed packages we index all symbols in user files. Symbols from user files won't appear in completions unless they are already used, otherwise the large number of symbols would make the completion list unmanageable. 
```

debonte, line 250: > are already used

Already used in the current file? An open file?

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+But since indexing is an expensive operation, we do not index every symbol from all installed packages but only top-level symbols. In another words, symbols in ```__all__``` in ```package/__init__```. If users want more than that, they need to use ```python.analysis.packageIndexDepths``` to tell us which packages they want us to spend more time on. Symbols indexed will show up both in completion and code actions.
+
+For user files, we will index up to 2000 files. One difference between user files and installed packages is that symbols from user files won't appear in completion unless they are already used. Unlike installed packages, we, by default, index all symbols from user files, which can cause thousands of user symbols to show up in completion. 
+
+We show symbols from user files only in the code action since it requires an explicit user action ```Ctrl+.``` rather than being automatically triggered like completion.
```
debonte, line 252: Is this saying that with `Ctrl+.` you get the full list of user symbols? Or that you get the user symbols that are "already used"?

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
...
+
+For people who don't want ```auto import``` suggestions to ever show up in completions since they feel it bloats the completions, you can use ```python.analysis.autoImportCompletions``` to disable it. Code actions can still be used with this setting disabled..
```
debonte, line 254: Extra period
```suggestion
For people who don't want ```auto import``` suggestions to ever show up in completions since they feel it bloats the completions, you can use ```python.analysis.autoImportCompletions``` to disable it. Code actions can still be used with this setting disabled.
```

README.md
```diff
@@ -223,6 +237,22 @@ Example of customizing semantic colors in settings.json:
 
     -   Convert import format according to `python.analysis.importFormat`.
 
+# Auto/Add import and indexing
+
+Pylance has a feature called ```auto import``` which shows symbols from installed packages and user files during completion. When committed, pylance will automatically generate an ```import``` statement. It also has something called ```add import``` which shows symbols from installed packages and user files for an unknown identifier. When selected, this will also insert an ```import``` statement.
```
debonte, line 242: Is `add import` the "code action" that you refer to below? If so, that wasn't clear to me. Maybe include the words "code action" in this paragraph as a way of introducing the term?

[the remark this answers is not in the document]
debonte: Maybe a link to VS Code docs would help? https://code.visualstudio.com/docs/editor/intellisense

README.md
debonte: Can we clarify that here? If it's too complex to explain here, maybe point the user to the detailed section below?

README.md
debonte: "Used to override the depth of installed package indexing on a per package basis. By default, only top-level modules are indexed (depth = 1). To index submodules, increase depth by 1 for each level of submodule you want to index."

I'm not totally satisfied with this. Maybe someone else can improve it. :)

luabud: I actually like it a lot. I think the example below helps illustrate what this means.

luabud:
One thought I just had (and please don't hate me for not thinking about this earlier 😅) is to move this whole section here to our docs, right above "Troubleshooting IntelliSense" section: https://code.visualstudio.com/docs/python/editing#_troubleshooting-intellisense. 

The code for it is here:  https://github.com/microsoft/vscode-docs/blob/main/docs/python/editing.md#troubleshooting-intellisense.

README.md
luabud: We could do `include all symbols?` to reflect it's a bool, and then add one more bullet point under the accepted values explaining "If `include all symbols` is set to false, only symbols in `__all__` are included. When it's set to `true`, Pylance will index every module/top level symbol declarations in the file".

heejaechang:
@luabud do you want to continue on here or should I close this PR?

luabud:
Let's close it!

PR #3388, opened 2022-09-24
Update LICENSE

mit-net-ai:


debonte (approved):
Wow, good eye. :) Thanks for the fix.

I compared it against the current version of Creative Commons 4.0 International and noticed an underscore that shouldn't have been there, so I added that into your PR.

debonte:
Oh, wait, you were removing the period? I misread that.

PR #3427, opened 2022-10-03
Automate scrum board Triage column

debonte:
For https://github.com/microsoft/pyrx/issues/2802

Attempt at automatically adding new issues into the Triage column of the scrum board.

FYI, the `microsoft` org admins have [provided](https://github.com/microsoft/github-operations/issues/47) an org-level PAT named `PROJECT_BOARD_ACTIONS_TOKEN`. So we should be able to test this out now.

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -10,6 +10,13 @@ jobs:
   triage-labels:
     runs-on: ubuntu-latest
     steps:
+      - name: If no labels, add to scrum board
+        uses: peter-evans/create-or-update-project-card@v2
+        if: ${{ join(github.event.issue.labels.*.name, ',') == '' }
+        with:
+          project-name: Python
+          column-name: Triage
+
       - name: If no labels, add triage
         id: no-labels
         uses: andymckay/labeler@master
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -14,7 +14,9 @@ jobs:
         uses: peter-evans/create-or-update-project-card@v2
         if: ${{ join(github.event.issue.labels.*.name, ',') == '' }}
         with:
-          project-name: Python
+          token: ${{ secrets.PYLANCE_BOT_GITHUB_TOKEN }}
+          project-location: microsoft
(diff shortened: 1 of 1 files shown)


.github/workflows/issues.yml
```diff
@@ -10,6 +10,13 @@ jobs:
   triage-labels:
     runs-on: ubuntu-latest
     steps:
+      - name: If no labels, add to scrum board
+        uses: actions/add-to-project@v0.3.0
+        if: ${{ join(github.event.issue.labels.*.name, ',') == '' }}
+        with:
+          github-token: ${{ secrets.PROJECT_BOARD_ACTIONS_TOKEN }}
+          project-url: https://github.com/orgs/microsoft/projects/145/views/22
```
judej, line 18: I think it should be added to view 1. The view 22 is this month's view and will deleted

debonte: Oh, or maybe view 17? https://github.com/orgs/microsoft/projects/145/views/17

.github/workflows/issues.yml
```diff
@@ -10,6 +10,13 @@ jobs:
   triage-labels:
     runs-on: ubuntu-latest
     steps:
+      - name: If no labels, add to scrum board
+        uses: actions/add-to-project@v0.3.0
+        if: ${{ join(github.event.issue.labels.*.name, ',') == '' }}
```
judej, line 15: Not sure about the workflow: Will this trigger if any time in the lifetime of the issue, there are no labels?
Should we check for `is:open`
Also, what if it is transferred? 

I think it is a good start.

.github/workflows/issues.yml
debonte: I was a bit unclear on how columns were handled -- how to get the new items to appear in the board's triage column. Maybe using view 17 is the solution there.

.github/workflows/issues.yml
debonte: What if I change it to add to the board if the "triage-needed" label is added?

judej (approved):
lgtm

Issue #3407, opened 2022-09-29
User documentation on import hook-based editable installs (PEP 660)


debonte:
Add user documentation somewhere explaining:

- Why import hooks are a problem for static analysis tools
- The downside of opting out of import hooks -- your editable install may not match your real install as closely as it could.
- How to out-out of import hooks in common build backends. Graham started that [here](https://github.com/microsoft/pylance-release/wiki/FAQ#packages-are-not-found-when-using-editable-installs). We might need to move it for discoverability.
    - setuptools: [legacy behavior](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior) / [strict mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#strict-editable-installs)
    - [hatchling](https://hatch.pypa.io/latest/config/build/#dev-mode)
    - [pdm](https://pdm.fming.dev/latest/pyproject/build/#editable-build-backend)

For https://github.com/microsoft/pyrx/issues/2715

ofek:
Hatchling does the `.pth` file way by default

debonte:
> Hatchling does the `.pth` file way by default

Thanks @ofek. For hatchling I think we would just explain that `dev-mode-exact = true` enables import hooks and could cause issues with Pylance.

PR #3440, opened 2022-10-07
Docs on import hook-based editable installs (PEP 660)

debonte:
Fixes https://github.com/microsoft/pylance-release/issues/3407

Added details in `TROUBLESHOOTING.md` on why import hook-based editable installs cause problems for Pylance/Pyright and how to configure some common build backends to use `.pth` files instead.

@erictraut, should we replicate this info in https://github.com/microsoft/pyright/blob/main/docs/import-resolution.md? Or have a link between the docs in the two repos? Or?

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -39,11 +39,11 @@ be taken as relative to the workspace root.
 
 Note that if you are coming to Pylance from using the Microsoft Python Language Server, this setting has changed from `python.autoComplete.extraPaths` to `python.analysis.extraPaths`.
 
-### Editable install packages not found
+### Editable install modules not found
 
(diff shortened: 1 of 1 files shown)


rchiodo (approved):

TROUBLESHOOTING.md
```diff
@@ -39,6 +39,38 @@ be taken as relative to the workspace root.
...
+[PEP 660](https://peps.python.org/pep-0660/) enables build backends (ex. setuptools) to
+use import hooks to direct the [import machinery](https://docs.python.org/3/reference/import.html)
+to the package's source files rather than using a `.pth` file. Import hooks can provide
+an editable installation that is a more accurate representation of your real installation.
+However, because resolving module locations using an import hook requires executing Python
+code, they are not usable by Pylance, Pyright, and other static analysis tools. Therefore,
+if your editable install is configured to use import hooks, Pylance/Pyright will be unable
+to find the corresponding source files.
+
+If you want to use static analysis tools and you are willing to accept a lower-fidelity
```
erictraut, line 53: I recommend deleting "and you are willing to accept a lower-fidelity editable install". There really isn't a choice here. If you want to use static analysis tools, you will need to use ".pth" files. (I'm also skeptical of the assertion that the fidelity difference is meaningful to all but a tiny percentage of users who perform editable installs, so I don't think we should emphasize this point.)

erictraut (commented):
Looks good. I added one minor suggestion.

I also like the idea of replicating this in the pyright documentation.

PR #3468, opened 2022-10-12
updated changelog.

heejaechang:


rchiodo (approved):

README.md
```diff
@@ -121,6 +113,36 @@ Pylance provides users with the ability to customize their Python language suppo
         -   `true` (default)
         -   `false`
 
+-   `python.analysis.indexing`
+
+    - Used to specify whether Pylance should index installed third party libraries and user files to provide features such as auto-import, add import, workspace symbols, etc.
+    - Available values:
+        - `null` (default)
```
debonte, line 120: I realize this isn't new (just moved), but what does null mean? What is the resulting behavior? Is it equivalent to true?

README.md
```diff
@@ -121,27 +113,57 @@ Pylance provides users with the ability to customize their Python language suppo
...
+            "includeAllSymbols": "whether to include all symbols (bool)"
+        }
+        ```
+        If `include all symbols` is set to `false`, only symbols in each package's `__all__` are included. When it's set to `true`, Pylance will index every module/top level symbol declarations in the file.  
+
+    -   Example:
+        ```jsonc
+        [
+            { "name": "sklearn", "depth": 2, "includeAllSymbols": true },
+            { "name": "matplotlib", "depth": 3, "includeAllSymbols": false }
```
bschnurr, line 142: if name is "" ie. empty does that override the default behaviour?

PR #3805, opened 2023-01-04
Adopt jupyter assign issues rotation

rchiodo:
Auto assign issues based on day of the week


rchiodo:
Current rotation:

Saturday: Bill
Sunday: Erik
Monday: Heejae
Tuesday: Jude
Wednesday: Rich
Thursday: Rich
Friday: Erik

At standup we talked about how to make triage go faster. The proposed solution was to auto assign a person every day to be responsible for the first level of investigation of a bug. 

That person would
- Determine if there is a bug or not
- Do some root cause analysis if possible
- Assign to triage if something needs to be decided on what to do

heejaechang (approved):

PR #4040, opened 2023-03-04
Minor tweaks to settings reference

luabud:
Minor edits from things I noticed when updating our official settings docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,9 +116,8 @@ Pylance provides users with the ability to customize their Python language suppo
 -   `python.analysis.indexing`
 
     - Used to specify whether Pylance should index installed third party libraries and user files to provide features such as auto-import, add import, workspace symbols, etc.
-    - Available values:
-        - `null` (default)
-        - `true`
+    - Accepted values:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -189,26 +188,26 @@ Pylance provides users with the ability to customize their Python language suppo
         -   `true`
         -   `false` (default)
 
-```python
-@pytest.fixture()
-def my_fixture() -> str:
-    return "foo"
+    ```python
```
rchiodo, line 191: Did you preview this change? I had to mess with a bunch to get github to render this correctly.

rchiodo (approved):

README.md
debonte: > build the extension and see if it renders appropriately?

This should be unrelated to what the user sees in VS Code, right? I believe that text comes from https://github.com/microsoft/pyrx/blob/main/packages/vscode-pylance/README.md.

rchiodo: No I think the preview renderer was faithful to what github did. Thanks for checking.

PR #4054, opened 2023-03-09
Create stale.yml

StellaHuang95:
Also add the bot to pylance repo to close "waiting for user response" issues after 30 idle days.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,34 @@
+# This workflow closes issues that have been labeled as 'waiting for user response' and had no activity for 30 days.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Close stale 'waiting for user response' issues
+
+on:
+  schedule:
+    #        ┌───────────── minute (0 - 59)
+    #        │ ┌───────────── hour (0 - 23)
+    #        │ │ ┌───────────── day of the month (1 - 31)
+    #        │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+    #        │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+    #        │ │ │ │ │
+    #        │ │ │ │ │
+    #        │ │ │ │ │
+    #        * * * * *
+    - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+      - uses: actions/stale@v5
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-issue-stale: 30
+          days-before-issue-close: 0
+          close-issue-message: "This issue was closed because it has been stalled for 30 days with no activity. If the issue still persists, please reopen with the information requested. Thanks."
+          any-of-labels: "waiting for user response"


.github/workflows/stale.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+      - uses: actions/stale@v5
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-issue-stale: 30
+          days-before-issue-close: 0
```
debonte, line 32: Since this is automated, I think it would be worth enabling the `stale-issue-message` to remind users to provide the missing info before we give up on them.

Here's the string that [vscode](https://github.com/microsoft/vscode/blob/main/.github/workflows/needs-more-info-closer.yml#L30) uses: "Hey @${assignee}, this issue might need further attention.\n\n@${author}, you can help us out by closing this issue if the problem no longer exists, or adding more information."

rchiodo (approved):

.github/workflows/stale.yml
StellaHuang95: Thanks, just added that.

debonte: I forgot to mention that vsode is using their own custom action for this. I just checked and it appears that the stale action doesn't support `author` and `assignee` replacement strings. See https://github.com/actions/stale/issues/177.

It's not clear to me where the implementation of vscode's action is. Maybe we can leverage it somehow though?

If not, how about a more generic string like this?:
"You can help us out by closing this issue if the problem no longer exists, or by adding more information. Otherwise, this issue will be closed in 1 week."

StellaHuang95: I could be wrong, but by reading the [VS Code's Issue Triage GitHub Actions](https://github.com/microsoft/vscode-github-triage-actions), it seems to only closes issues that are labeled with [needs more info](https://github.com/microsoft/vscode-github-triage-actions#:~:text=issues%20that%20are%20marked%20a%20needs%20more%20info%20label) after certain time, which we are not using.

Another approach I tried is to add another workflow that looks similar to this:
```yaml
on:
  issues:
    types:
      - labeled
  workflow_dispatch:

jobs:
  add-comment:
    if: github.event.label.name == 'Stale'
    runs-on: ubuntu-latest
    permissions:
      issues: write
    steps:
      - name: Add assignee and author tags to issue comment
        uses: peter-evans/create-or-update-comment@v2
        with:
          issue-number: ${{ github.event.issue.number }}
          body: "This issue has been labeled with ${{github.event.label.name}} by @${{github.event.sender.login}}. Assignee is @${{github.event.issue.assignee.login}} and author is @${{github.event.issue.user.login}}."
```
But that also means the workflow will run every time a label is added to the issue, not sure if that would be overkill. And for some reason, this workflow is only triggered by labels added manually, so I just added a more generic string to it.

debonte: > I could be wrong, but by reading the [VS Code's Issue Triage GitHub Actions](https://github.com/microsoft/vscode-github-triage-actions), it seems to only closes issues that are labeled with [needs more info](https://github.com/microsoft/vscode-github-triage-actions#:~:text=issues%20that%20are%20marked%20a%20needs%20more%20info%20label) after certain time, which we are not using.

I'm not sure I understand what you're saying. I think you're either saying that vscode is using a different "needs info" label than we are or that they don't support pinging the user for more info.

But the label name can be configured on vscode's action via `label` and pinging for more info can be enabled via `pingDays` and `pingComment`. 

If neither of those are what you were getting at, can you rephrase the question?

> Another approach I tried is to add another workflow that looks similar to this:

Oh, that seems good to me. I wouldn't worry about this being overkill. We already have a set of actions that run when issues are created/labeled/unlabeled. See https://github.com/microsoft/pyrx/blob/main/.github/workflows/issues.yml

StellaHuang95: Oh I think you're right! I got confused. I thought the label has to be exactly `needs more info` to be able to be closed by the bot. I will give it another try.😊

.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Checkout Actions
+        uses: actions/checkout@v3
+        with:
+          repository: "microsoft/vscode-github-triage-actions"
+          path: ./actions
+          ref: stable
+      - name: Install Actions
+        run: npm install --production --prefix ./actions
+      - name: Run Stale Issue Closer
+        uses: ./actions/needs-more-info-closer
```
rchiodo, line 31: Doesn't it need this action?

rchiodo: I assume that action would be in the actions folder

rchiodo (changes requested):
I think you might be missing the action

.github/workflows/stale.yml
StellaHuang95: Hi @debonte, I've changed to use vscode's action. I have tested on my own repo, but I can also add the `workflow_dispatch` event if you want to manually trigger the workflow and test it:)

.github/workflows/stale.yml
rchiodo: oh yeah, you probably need to copy the source here:
https://github.com/microsoft/vscode-github-triage-actions/tree/stable/needs-more-info-closer

StellaHuang95: So you're saying I need to create an action folder in our repo and also copy the source to our repo? I am new to git actions so maybe I misunderstood something.🥲My understanding is that the `Checkout Actions` step checks out the microsoft/vscode-github-triage-actions repository at the stable branch to the ./actions directory within the current repository. Then the `Install Actions` step installs some dependencies in the ./actions directory. And the last step runs the local action file located at ./actions/needs-more-info-closer

rchiodo: Oh sorry, I didn't notice that code. That sounds like it could work too.

rchiodo (approved):

.github/workflows/stale.yml
debonte: Nice! :)

debonte (approved):

Issue #4163, opened 2023-03-29
Pyright and pylance disagree


FrancescElies:
## Environment data

-   Language Server version: Pylance language server 2023.3.30 (pyright 68a8e98a) and `pyright -> 1.1.301`
-   OS and version: windows 11 (22621)
-   Python version (& distribution if applicable, e.g. Anaconda): `3.11.1`

## Code Snippet

![image](https://user-images.githubusercontent.com/6558089/228525176-4e26f900-4968-4ebb-9c0f-a563ffa3145d.png)
`pyright -> 1.1.301` in CI and over the cli don't complain
`matplotlib -> 3.7.1`
`pylance -> v2023.3.30`

## Additional info
Other editors (e.g. helix) configured with pyright as lsp server are in sync with pyright's view of the world and don't have this issue, that's why I suspect something going on in _vscode_ with  _pylance_

I also think this issue has nothing to do with the latest release pyright, we have seen this a couple of times in the past  weeks with older versions of pyright too.
## Repro Steps
See [here](https://github.com/microsoft/pylance-release/issues/4163#issuecomment-1489917429)

## Expected behavior

Pyright when called over the cli and editing code in vscode (pylance) report the same issues

## Actual behavior

Pyright and pylance disagree

## Logs

```
TODO: do upon request if needed.
```

erictraut:
@judej, can we please prioritize [this issue](https://github.com/microsoft/pylance-release/issues/4085)? Documentation is badly needed here.

FrancescElies:
ℹ️ On our side we are working on a reprex, if we are lucky by tomorrow we might be able to provide it

FrancescElies:
Current workaround for us was to set `reportUnnecessaryTypeIgnoreComment = false`

erictraut:
@FrancescElies, it looks like you're not setting `useLibraryCodeForTypes` in your pyproject.toml. If left unspecified, this setting defaults to `false` in pyright but `true` in pylance. I recommend that you explicitly set it to `false` in your pyproject.toml.

rchiodo:
I'm currently working on the documentation. Hope to have something for review today.

rchiodo:
See the linked PR. Comments welcome

FrancescElies:
[...]

Issue #4085, opened 2023-03-16
Create documentation on the topic of using pylance and pyright (CLI) together


judej:
Users are seeing inconsistencies in errors and warnings when using VSC, VS, Pyright. 

Related issues:
https://github.com/microsoft/pylance-release/issues/3045
https://github.com/microsoft/pylance-release/discussions/3638
https://github.com/microsoft/pylance-release/issues/3838
https://github.com/microsoft/pyright/issues/4780
https://github.com/microsoft/pyright/issues/4784

blakeNaccarato:
Thanks for putting this on the roadmap! Documenting the details of Pylance's type stub resolution will go a long way to help users align the behavior of vanilla `pyright` with Pylance, for instance in CI like GitHub Actions. Currently, the audiences that might benefit from this documentation are in camps (1) and (2) below, with (3) and (4) being extensions of camp (2) requiring feature development either in Pylance or VSCode.

1. Use `mypy` for type-checking, locally and in CI, and only use Pylance for refactoring features and QoL in your IDE.
2. Use `pyright` for type-checking in CI. Locally, stay at latest Pylance, and be okay with troubleshooting minor discrepancies with CI. Check in some stubs to the repo to reduce the discrepancies.
3. Desired feature: The same as (2), but allow Pylance to be configured to not use bundled stubs at all, giving full control to the developer to supply their stubs in their repo's `typings` folder.
4. Ridiculously complex solution that should not be considered (but mentioned here for completeness): The same as (2), but VSCode enables workspace-specific extension versioning granularity. Then, Pylance can be manually synchronized with the hypothetically-perfect stubs in the repo's `typings`.

I suspect that most people are in camp (1), but sufficient documentation of this thorny issue would make camp (2) more viable, especially for bigger projects. I also hope that the documentation addresses some of Pylance's type stub resolution mysteries, such as:

-  Pylance bundles stubs at `~/.vscode/extensions/ms-python.vscode-pylance-<RELEASE>/dist`. Does `typeshed-fallback` factor into stub resolution at all, and give rise to any possible pyright/Pylance discrepancies?
- The subfolder `bundled` has `native-stubs` and `stubs` directories. There is some overlap in their contents. Which stubs take precedence, and when? Are they loaded in a certain order?
- Is it possible to map the stub-loading logic of Pylance to an equivalent singular `typings` folder containing a certain set of stubs? If so, how might that be composed from the above stubs?
[...]

Issue #3857, opened 2023-01-23
`region` at the start of a normal comment triggers error about `endregion` being missing


rchl:
## Environment data

-   Language Server version: v2023.1.30
-   OS and version: macOS
-   Python version (& distribution if applicable, e.g. Anaconda): -

## Code Snippet

```python
            for diagnostic, candidate in diagnostics:
                # Checking against points is inclusive unlike checking whether region intersects another
                # region which is exclusive (at region end) and we want an inclusive behavior in this case.
```

## Repro Steps

1. Have a normal comment in the file that starts with a "region" word

## Expected behavior

No error about missing "endregion" pair

## Actual behavior

Getting error about "endregion" missing

## Logs

Not relevant

heejaechang:
@debonte dup?

debonte:
Not a dupe. This is really about multi-line comments that happen to have one line starting with "region".

@rchl, thanks for filing this.

debonte:
I don't see a way to improve detection of `#region` comments that would help in this scenario without breaking more common region scenarios and also making Pylance behave significantly differently from other Python editing experiences (ex. PyCharm). For example:
1. Require some specific formatting (ex. `#region:` with a colon). 
2. Assume that `#region` at the start of a line other than the first line in a multi-line comment is not intended to be a region. 

I think the best solution is to provide a way to disable the unmatched region diagnostics. This seems especially desirable in Pyright which doesn't support region folding. At least Pylance users have folding support as an incentive to get their `#region` comments recognized correctly.

@erictraut, any concerns with adding a new `DiagnosticRule` for this? Maybe disabled by default in Pyright and on by default in Pylance? I think Rich added a mechanism for `reportShadowedImports` that makes that possible.

@rchl, would you be satisfied with this solution?

Workarounds in the meantime to silence a particular instance of this diagnostic:
1. Rejustify the text in the comment so "region" isn't at the start of a line.
2. Add an extra `#` before the line in question (`## region`).
1. `#type: ignore`, but that seems ridiculous in the midst of a multi-line comment/paragraph.

erictraut:
This doesn't strike me as something that justifies a new diagnostic rule. If we were to add a new diagnostic rule, it should definitely not have a different default because pyright and pylance. We don't have any precedent for that. The defaults are determined by the "none", "basic" or "strict" rule sets.

[...]

PR #4172, opened 2023-03-30
Add a document describing differences between Pylance and Pyright settings

rchiodo:
Fixes https://github.com/microsoft/pylance-release/issues/4163
Fixes https://github.com/microsoft/pylance-release/issues/4085

Still need a place for users to find all of our bundled stubs.


USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+### Diagnostic Severity Overrides to match Pyright defaults
+
+Here's the settings you would provide to Pylance based on which Pyright mode you are using:
+
+<details>
+  <summary>Pyright 'off' mode</summary>
+
+```json
+{
+    "python.anaylsis.diagnosticSeverityOverrides": {
```
debonte, line 61: Could we limit this to just the diagnostics which differ? I thought `reportShadowedImports` was the only one that was different? But maybe I'm misunderstanding what issue this is addressing.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
+# Pyright and Pylance Together
+
+[Pyright](https://github.com/Microsoft/pyright) is an open source version of a Python type checker that forms the basis for Pylance. Both Pyright and Pylance can be used inside of an IDE to provide a [language server](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#why-language-server) for Python code.
+
+Pyright can also be used on the [command line](https://github.com/microsoft/pyright/blob/main/docs/command-line.md) to produce json or text output. This can be used in a variety of situations, like a pre-commit hook, or a [github action](https://github.com/jakebailey/pyright-action).
+
+However, Pylance and Pyright don't always produce the same results. The reason for this is a difference in configuration. 
+
+## Stub differences
```
erictraut, line 10: I recommend using the term "Type stub" at least once before using the more informal term "stub".

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
+# Pyright and Pylance Together
+
+[Pyright](https://github.com/Microsoft/pyright) is an open source version of a Python type checker that forms the basis for Pylance. Both Pyright and Pylance can be used inside of an IDE to provide a [language server](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#why-language-server) for Python code.
+
+Pyright can also be used on the [command line](https://github.com/microsoft/pyright/blob/main/docs/command-line.md) to produce json or text output. This can be used in a variety of situations, like a pre-commit hook, or a [github action](https://github.com/jakebailey/pyright-action).
+
+However, Pylance and Pyright don't always produce the same results. The reason for this is a difference in configuration.
```
erictraut, line 8: There are multiple reasons including configuration differences and bundled stubs. Perhaps it's best to just say "There are several reasons why such differences may be observed." (or something to that effect).

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+## Stub differences
+
+Pylance comes bundled with a bunch of stubs, usually found here in the installation:
+
+```
+<pylance-extension-install>/dist/bundled
+<pylance-extension-install>/dist/native-stubs
+```
+
+In order for Pyright to produce the same errors as Pylance it need to use the same stub files.
```
erictraut, line 19: Grammar nit: Delete "In order".

erictraut, line 19: Another grammar nit: Include comma before "it needs to use". Also, "need" should be "needs".

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+Pylance comes bundled with a bunch of stubs, usually found here in the installation:
+
+```
+<pylance-extension-install>/dist/bundled
+<pylance-extension-install>/dist/native-stubs
+```
+
+In order for Pyright to produce the same errors as Pylance it need to use the same stub files. 
+
+This might be done by copying the contents of each of those folders to a new folder.
```
erictraut, line 21: This sentence is pretty vague. Copying what folders into what new folder?

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+
+In order for Pyright to produce the same errors as Pylance it need to use the same stub files. 
+
+This might be done by copying the contents of each of those folders to a new folder. 
+
+You'd then reference that new folder in your [pyrightconfig.json](https://microsoft.github.io/pyright/#/configuration)
+
+```json
+{
+    "stubPath": "./pylanceStubs"
```
erictraut, line 27: It might be easier just to recommend that they place the stubs in a project-local "typings" directory. That way, no additional configuration is needed because pyright and pylance will both pick them up from that location.

If you retain the configuration section, then I recommend also including an example of how to configure it in `pyproject.toml`. I think that more pyright users use `pyproject.toml` than `pyrightconfig.json` nowdays.

USING_WITH_PYRIGHT.md
rchiodo: No a bunch of them differ. And it was easier to just list this out than try and keep track of the differences.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+}
+```
+
+Where `pylanceStubs` contains the contents of the `bundled` and `native-stubs` folders.
+
+We have plans to make this easier in the [future](https://github.com/microsoft/pylance-release/discussions/3638).
+
+## Settings differences
+
+Both Pylance and Pyright have [settings](https://github.com/microsoft/pylance-release#settings-and-customization) to control behavior. For the most part, they use the same values, but the few differences do have an impact on the analysis.
```
erictraut: It might be worth talking about "language server settings" versus "pyright configuration settings". The latter are found in a `pyrightconfig.json` or `pyproject.toml` file and tend to override the former, although there are cases where the two are composed together. (It's complicated.) In general, if users want consistent behavior between pyright and pylance, I recommend that they do not use language server (e.g. VS Code) settings and instead set all of their configuration settings in the `pyrightconfig.json` or `pyproject.toml` file.

If users follow this recommendation, then you can remove most of the text below because pyright and pylance will both pick up these settings and act consistently between the two.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+
+| Setting | Pylance default | Pyright default | Description | Potential Impact |
+|----|----|----|----|----|
+| python.analysis.autoSearchPaths | true | false|  Adds 'src' to the list of search paths | This may change what files are found when analyzing. So if you're getting missing imports for things in your 'src' tree, this might be why |
```
erictraut: I think you can delete the section on "python.analysis.autoSearchPaths". I'm pretty sure that pyright and pylance use the same default value here, so if you don't override the default in your language server settings, you will get the same behaviors.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+| python.analysis.indexing | true | false | All files in the current virtual environment are parsed to a certain depth. This gives Pylance more information for auto import completions and other things. | This can have an impact on completions and hover information, but shouldn't impact errors returned.
+| python.analysis.useLibraryCodeForTypes | true | false | When true, this means 3rd party libraries are analyzed to produce type information. Without it, all 3 party libraries are just assumed to be of type Any unless they provided type information explicitly | This setting changes the types found by Pylance/Pyright. It can definitely cause differences in errors reported, so make this the same for both if you want consistent results. A lot of the time this is the sole cause of differences. |
```
erictraut: I've been considering changing pyright's default for `useLibraryCodeForTypes` to true. This would be a breaking change for pyright users, but this would eliminate a common source of confusion for those who want to use both pyright and pylance. Thoughts?

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+
+[Pyright](https://github.com/Microsoft/pyright) is an open source version of a Python type checker that forms the basis for Pylance. Both Pyright and Pylance can be used inside of an IDE to provide a [language server](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#why-language-server) for Python code.
+
+Pyright can also be used on the [command line](https://github.com/microsoft/pyright/blob/main/docs/command-line.md) to produce json or text output. This can be used in a variety of situations, like a pre-commit hook, or a [github action](https://github.com/jakebailey/pyright-action).
+
+However, Pylance and Pyright don't always produce the same results. The reason for this is a difference in configuration. 
+
+## Stub differences
+
+Pylance comes bundled with a bunch of stubs, usually found here in the installation:
```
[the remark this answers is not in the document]
rchiodo: Whoops, will fix.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+| python.analysis.indexing | true | false | All files in the current virtual environment are parsed to a certain depth. This gives Pylance more information for auto import completions and other things. | This can have an impact on completions and hover information, but shouldn't impact errors returned.
+| python.analysis.useLibraryCodeForTypes | true | false | When true, this means 3rd party libraries are analyzed to produce type information. Without it, all 3 party libraries are just assumed to be of type Any unless they provided type information explicitly | This setting changes the types found by Pylance/Pyright. It can definitely cause differences in errors reported, so make this the same for both if you want consistent results. A lot of the time this is the sole cause of differences. |
+| python.analysis.typeCheckingMode | off | basic | Determines what diagnostics are shown. | Pylance defaults to off, but there is a possibility that VS code will force this to 'basic' for pylance users. If you want to guarantee this is the same as Pyright, force it to 'basic' (or whatever you want to enforce) by specifying it in your settings.json|
```
erictraut: I think you can delete the section on "python.analysis.typeCheckingMode". Now that pylance defaults to "basic", pyright and pylance use the same default value, so if you don't override the default in your language server settings, you will get the same behaviors.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+| python.analysis.diagnosticMode | openFilesOnly | openFilesOnly | Determines which files are analyzed. `openFilesOnly` means only those files that are open in an IDE | You'll likely set this to 'workspace' for the CI version of Pyright. This means Pylance would likely not give you all of the same errors since it's only looking at open files. |
```
erictraut: The diagnosticMode has no bearing on which diagnostics are produced. It applies only to language servers, not the CLI. I recommend removing this section because it will cause confusion.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+| python.analysis.diagnosticSeverityOverrides | - | - | Allows a user to override the severity levels for individual diagnostic rules. | See below |
```
erictraut: I think you can delete the section on "python.analysis.diagnosticSeverityOverrides". Pyright and pylance use the same default value here, so if you don't override the default in your language server settings, you will get the same behaviors.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Diagnostic severity overrides
+
+Both Pylance and Pyright specify a custom [`python.analysis.diagnosticSeverityOverrides`](https://microsoft.github.io/pyright/#/configuration?id=diagnostic-rule-defaults) setting. Differences in this setting have a direct impact on the diagnostics returned. If you were attempting to get the same error output for both, making sure this setting is consistent is probably the first thing to do.
+
+
+### Diagnostic Severity Overrides to match Pyright defaults
```
erictraut, line 56: I recommend deleting this section because these defaults are the same between pyright and pylance. Including it here will confuse users more than help. If, for some reason, you think that it's important to explain the default values, I recommend simply providing a link to [pyright's documentation](https://microsoft.github.io/pyright/#/configuration?id=diagnostic-rule-defaults) rather than duplicating all of this information here. It will inevitably become stale as we add more diagnostic rules.

USING_WITH_PYRIGHT.md
erictraut: Oh, I forgot about `reportShadowedImports`. That one does differ between pyright and pylance, doesn't it? 

If so, we should consider removing this difference. If we decide to leave the difference, then I agree with Erik that we should simply document the difference rather than repeating all of the default values that do not differ.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,305 @@
...
+        "reportUnnecessaryTypeIgnoreComment": "none",
+        "reportUnusedCallResult": "none",
+    }
+}
+```
+
+</details>
+</br>
+
+You would add these settings to your settings.json when running inside of VS Code.
```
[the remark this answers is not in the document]
erictraut: For users who are using both pyright and pylance, I recommend against using VS Code settings. They should instead use `pyrightconfig.json` or `pyproject.toml`. As I mentioned above, this guarantees that they will get consistent results.

USING_WITH_PYRIGHT.md
debonte: Weird. Based on these comments from Eric I thought it was just that one. Maybe we should consider getting these back in sync?

https://github.com/microsoft/pylance-release/issues/3857#issuecomment-1407167773
https://github.com/microsoft/pylance-release/issues/3857#issuecomment-1407208449

rchiodo: Maybe this was the wrong spot to look, but if you compare the default values in https://github.com/microsoft/pyright/blob/main/packages/vscode-pyright/package.json with the values in https://github.com/microsoft/pyrx/blob/main/packages/vscode-pylance/package.json, they differ in like 12 spots

erictraut: Hmm, that's the documentation, not the source of truth. I think if you look at the source of truth in the code (https://github.com/microsoft/pyright/blob/09c79ed3e284b953b4d96c240a2dca91d74b2eee/packages/pyright-internal/src/common/configOptions.ts#L421), I think (hope?) that you'll see more consistency. If you're finding many differences in the documentation, we should fix those as well to ensure that they're consistent with the behaviors defined in the code.

rchiodo: Okay I'll enter a bug to fix the documentation in the pylance extension. Althought it's kind of ambiguous there too because the default type checking mode is off, unless you're in an experiment, then it's basic. Which changes the default value for a bunch of settings.

erictraut: I thought that experiment ended a long time ago and now everyone was opted in to "basic" by default. Is that not the case? If not, why is the experiment still going after more than a year? What additional information is needed to make a decision?

rchiodo: Not sure myself, @luabud do you know if this is at 100% or not? Perhaps we can just eliminate the experiment now?

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,73 @@
...
+| python.analysis.indexing | true | false | All files in the current virtual environment are parsed to a certain depth. This gives Pylance more information for features such as auto import. | This can have an impact on completions and hover information, but shouldn't impact errors returned.|
+| python.analysis.useLibraryCodeForTypes | true | false | When true, 3rd party libraries are analyzed to produce type information. Without it, all 3rd party library types are assumed to be of type `Any` unless they provided type information explicitly. | This setting changes the types found by Pylance/Pyright and therefore can cause differences in errors reported, so make this the same for both if you want consistent results. A lot of the time this is the sole cause of differences. |
+| python.analysis.typeCheckingMode | off | basic | Determines what diagnostics are shown. | Pylance defaults to `off`, but there is a possibility that VS code will force this to `basic` for Pylance users. If you want to guarantee this is the same as Pyright, set it to `basic` (or whatever you want to enforce) by specifying it in your settings.json. |
```
luabud, line 52:
```suggestion
| python.analysis.typeCheckingMode | off | basic | Determines what diagnostics are shown. | Pylance defaults to `off`, but there is a possibility that VS code will default this to `basic` for Pylance users. If you want to guarantee this is the same as Pyright, set it to `basic` (or whatever you want to enforce) by specifying it in your settings.json. |

```

luabud: force might be a strong word. It defaults to basic in VS Code insiders and to off in VS Code stable.

USING_WITH_PYRIGHT.md
rchiodo: This looks like the default is 'false' for pyright command line?
https://github.com/microsoft/pyright/blob/09c79ed3e284b953b4d96c240a2dca91d74b2eee/packages/pyright-internal/src/analyzer/service.ts#L699

Maybe I'm not following the code flow correctly here.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Settings differences
+
+Both Pylance and Pyright have [settings](https://github.com/microsoft/pylance-release#settings-and-customization) to control behavior. For the most part, they use the same values, but the few differences, described below, do have an impact on the analysis.
+
+
+| Setting | Pylance default | Pyright default | Description | Potential Impact |
+|----|----|----|----|----|
+| python.analysis.autoSearchPaths | true | false|  Adds 'src' to the list of search paths. | This may change what files are found when analyzing. So if you're getting missing imports for modules in your 'src' tree, this might be why. |
+| python.analysis.indexing | true | false | All files in the current virtual environment are parsed to a certain depth. This gives Pylance more information for features such as auto import. | This can have an impact on completions and hover information, but shouldn't impact errors returned.|
```
luabud, line 50:
```suggestion
| python.analysis.indexing | true | false | This setting enables all files in the current virtual environment to be parsed to a certain depth. This gives Pylance more information for features such as auto import. | This can have an impact on completions and hover information, but shouldn't impact errors returned.|

```

USING_WITH_PYRIGHT.md
rchiodo: I assume it makes analysis slower? Like a lot slower? But more accurate. I'd pick accuracy over speed.

USING_WITH_PYRIGHT.md
erictraut: I guess I was mistaken. Looks like this default is different too. Another one we should consider changing to make them the same. I'm fine switching the default to true for pyright if you don't see any problem with that.

USING_WITH_PYRIGHT.md
erictraut: I'm not so concerned about analysis speed. The pyright CLI is much less sensitive to speed than pylance because the former is used in a batch manner. There's no human impatiently waiting for immediate feedback. :) 

The issue is with accuracy of type information. The original thinking was that pyright needed accurate type information for type checking, whereas pylance could get away with somewhat inaccurate (inferred) type information because it was focused mostly on completion suggestions. But our type inference has gotten good enough now that I think it's reasonable to default `useLibraryCodeForTypes` to true for pyright. Users can always disable it if it's giving them too many inaccurate types.

USING_WITH_PYRIGHT.md
rchiodo: Well except that people using it will suddenly get subtle differences. I guess we have no good way to notify people of these sort of things other than release notes.

USING_WITH_PYRIGHT.md
rchiodo: According to what Luciana mentioned, we're are only doing 'basic' for insiders at the moment.

USING_WITH_PYRIGHT.md
erictraut: Yes, in theory a user could see a change in behavior, but I think it's unlikely to affect many (any?) pyright users. If a user has a "src" directory, they probably already configure it as an import search path using "extraPaths".

USING_WITH_PYRIGHT.md
rchiodo: Maybe I should add another entry to the table to specify the 'pyrightconfig.json' setting

rchiodo: I don't think people using the IDE will want a pyrightconfig.json that can also be impacted by settings.json? Not sure. Does pyrightconfig.json work in the .vscode folder or does it have to be in the root of the project?

USING_WITH_PYRIGHT.md
erictraut: Hmm. I think I'm correct in asserting that we've been running this experiment for the better part of a year. Presumably there was some question this experiment was trying to answer — probably to inform the decision to enable "basic" type checking in production? Do we not have enough data at this point to answer the question(s)? What additional data would be needed to finalize this decision?

USING_WITH_PYRIGHT.md
erictraut: The `pyrightconfig.json` or `pyproject.toml` must be in the root of the project. This is consistent with other tools (e.g. `tsconfig.json`). This file is meant to be checked in and shared across all project contributors, whereas the ".vscode/settings.json" file is often not checked in (since it contains per-user settings in many cases).

We should confirm my assumption that "there are cases where the two are composed together". Perhaps that's not the case and `pyrightconfig.json` and `pyproject.toml` configuration settings always prevail. I think there are cases where that's not true though. For example, if you specify a language server setting but don't specify the corresponding setting in the config file, I think the language server setting prevails (as opposed to the default value for that setting).

erictraut: In any case, what I say above is the advice I recommend to people who want the same behavior from pyright and pylance: don't use language server settings.

rchiodo: I tried the following:

- pyrightconfig.json = reportUndefinedVariables = none
- user settings.json = reportUndefinedVariables = error
- workspace settings.json = reportUndefinedVariables = error

pyrightconfig.json wins.
 
Recommending using pyrighconfig.json seems like a good idea because it will always trump everything else.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Settings differences
+
+Both Pylance and Pyright have [settings](https://github.com/microsoft/pylance-release#settings-and-customization) to control behavior. For the most part, they use the same values, but the few differences, described below, do have an impact on the analysis.
+
+
+| Setting | Pylance default | Pyright default | Description | Potential Impact |
+|----|----|----|----|----|
+| python.analysis.autoSearchPaths | true | false|  Adds 'src' to the list of search paths. | This may change what files are found when analyzing. So if you're getting missing imports for modules in your 'src' tree, this might be why. |
+| python.analysis.indexing | true | false | This setting enables all files in the current virtual environment to be parsed to a certain depth. This gives Pylance more information for features such as auto import. | This can have an impact on completions and hover information, but shouldn't impact errors returned.|
```
heejaechang, line 50: this option doesn't exist in pyright and this shouldn't impact type info. so I think we can omit it or have different section that shows pylance only or pyright only options.

USING_WITH_PYRIGHT.md
```diff
@@ -0,0 +1,74 @@
...
+```
+<pylance-extension-install>/dist/bundled
+<pylance-extension-install>/dist/native-stubs
+```
+
+For Pyright to produce the same errors as Pylance, it needs to use the same stub files. 
+
+This might be done by copying the contents of each of the bundled Pylance stub folders to a new folder. 
+
+You'd then reference that new folder in your [pyrightconfig.json](https://microsoft.github.io/pyright/#/configuration)
```
heejaechang, line 23: I think we want to list where got them?

https://github.com/microsoft/python-type-stubs and other ones we pull.

rchiodo: I think we'd handle that when we create an repository to make this easier. I mean I figure we'll just use git submodule and reference all of the locations that way.

rchiodo:
Does anybody have any more feedback? It seems pretty succinct to me now. After I submit, we can make more tweaks, but I think the general idea is good.

Issue #4232, opened 2023-04-12
Add a bot to remove "waiting for user response" label when the user replies.


StellaHuang95:
Per discussion, add a bot that automatically removes the "waiting for user response" label and adds a "user responded" label if the original poster replies.

PR #4257, opened 2023-04-19
Create auto-label.yml

StellaHuang95:
fixes https://github.com/microsoft/pylance-release/issues/4232

Add a bot that automatically removes the "waiting for user response" label and adds a "user responded" label if the original poster replies.

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,0 +1,41 @@
+name: Auto-label when user responds
+permissions:
+  issues: write
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  run-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add/Remove labels when user responds
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const payload = context.payload;
+            const issueAuthor = payload.issue.user.login;
+            const commentAuthor = payload.comment.user.login;
+            if (issueAuthor === commentAuthor) {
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-label.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  run-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add/Remove labels when user responds
+        uses: actions/github-script@v6
```
debonte, line 14: I think you could do this more simply with the [`andymckay/labeler`](https://github.com/andymckay/labeler) action that we [used to use](https://github.com/microsoft/pyrx/commit/9d9b1974dca8bcd2da16334adb98bbceab9dd768) for `triage-needed`.

```yml
      - name: Add/Remove labels when user responds
        uses: andymckay/labeler@master
        if ${{ github.event.comment.sender == github.event.issue.user && contains(github.event.issue.labels.*.name, 'waiting for user response') && !contains(github.event.issue.labels.*.name, 'user responded') }}
        with:
          add-labels: 'user responded'
          remove-labels: 'waiting for user response'
```

I'm not sure if the `user` comparison will work. You might need to compare `user.id`.

debonte (changes requested):

.github/workflows/auto-label.yml
StellaHuang95: @debonte Thanks! I changed to `github.event.comment.user.login == github.event.issue.user.login` and tested it in my personal repo, and it worked.

debonte (approved):

PR #4458, opened 2023-06-06
Update round robin to use 'updated' for sorting

rchiodo:
Since the validation only runs on an issue being created, 'updated' should list the last issue created whether it was new or transferred.

Fixes https://github.com/microsoft/pyrx/issues/3651

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -22,6 +22,7 @@ jobs:
             const issues = await github.rest.issues.listForRepo({
               state: 'all',
               assignee: '*',
+              sort: 'updated',
               owner: context.repo.owner,
               repo: context.repo.repo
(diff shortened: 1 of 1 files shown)


.github/workflows/issues.yml
```diff
@@ -22,6 +22,7 @@ jobs:
             const issues = await github.rest.issues.listForRepo({
               state: 'all',
               assignee: '*',
+              sort: 'updated',
```
debonte, line 25: Won't this mean that if I continually make changes to one of my existing issues, I won't get assigned any new issues?

rchiodo: Oh yeah you're right. I was thinking the issue coming in would be the one we'd want to look at. Duh.

PR #4578, opened 2023-07-05
Update issues.yml to handle removal of assignee field

rchiodo:
Fixes https://github.com/microsoft/pyrx/issues/3838


.github/workflows/issues.yml
```diff
@@ -24,7 +24,7 @@ jobs:
               owner: context.repo.owner,
               repo: context.repo.repo
             });
-            return issue.data.assignee.login || '';
```
rchiodo, line 27: assignee is coming back as 'null' now for some reason. Github may have updated their API.

PR #4618, opened 2023-07-17
Add 'needs repro' to issues as they come in

rchiodo:


rchiodo:
Tested locally to make sure it works using packages\pylance-internal\build\octokit.js

bschnurr (approved):

PR #5155, opened 2023-11-28
Adding some test files that can demo all existing features.

heejaechang:
these are some basic one to discuss and set pattern.


rchiodo (approved):
Looks great to me :)

StellaHuang95:
So what are the things that can't be caught by putting those tests in our fourslash/jest test suites?

heejaechang:
alright, then I will keep adding stuff to the test projects.

StellaHuang95 (approved):

heejaechang:
> So what are the things that can't be caught by putting those tests in our fourslash/jest test suites?

you can get all these things tested on our unit test and unit test should cover things in more detail. this is more sanity test (make sure things work in real vscode not just in unit test hardness) and these thing can also be automated using UI integration test (a test that uses vscode itself to test)

PR #5180, opened 2023-11-30
Covering workspace features

heejaechang:
cover all workspace features


testing/single/src/lib/userModule.py
```diff
@@ -5,4 +5,20 @@
 
 class MyType:
     def method(self, v: Path) -> ConnectTimeout:
-        raise Exception("Hello")
\ No newline at end of file
+        raise Exception("Hello")
```
rchiodo, line 8: Do you want to add comments what this file should be used for (or what its purpose is)?

PR #5187, opened 2023-11-30
basic notebook tests

heejaechang:
added basic notebook sanity check test files


rchiodo (approved):

testing/single/src/showCallHierarchy.py
```diff
@@ -48,8 +48,10 @@ def method(self):
 class BaseConsumer:
     def consumer(self, base: Base):
         base.method()
+        return base
```
StellaHuang95, line 51: Why is this added?

heejaechang: see `#test basic call hierarchy ...` cell.

PR #5340, opened 2024-01-09
Update USING_WITH_PYRIGHT to reflect Pyright's default typeCheckingMode now being `standard`

debonte:
Updated doc to say that Pyright defaults to `standard` now instead of `basic`.

> but there is a possibility that VS code will default this to `basic` for Pylance users

Also remove this sentence because I believe it was referring to the typeCheckingMode experiment which is not longer running. @luabud is that correct?


rchiodo (approved):

luabud (approved):
great catch, thank you! and yes that's correct, we no longer override this value in VSC

PR #5532, opened 2024-02-28
Update README.md

yilinjuang:
autoImportUserSymbols was removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,12 +154,6 @@ Pylance provides users with the ability to customize their Python language suppo
         - `true`
         - `false` (default)
 
-- `python.analysis.autoImportUserSymbols`
-    - Used to control the offering of user symbols in auto-import completions. This will impact number of items shown in the completion and performance.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,12 +154,6 @@ Pylance provides users with the ability to customize their Python language suppo
         - `true`
         - `false` (default)
 
-- `python.analysis.autoImportUserSymbols`
```
rchiodo, line 157: This may not be in the package.json but we still read this value in the code. @Heejaechang, I believe you added this for an experiment in this commit: https://github.com/microsoft/pyrx/commit/893f8cf8dd9763f9125a2ecd22b29f88b65a97a7

The `includeUserSymbolsInAutoImport` is still used internally? Should we remove it?

heejaechang: it has been there for long time and no one complained about it, so I think we can remove it.

rchiodo (approved):

PR #5701, opened 2024-03-28
Add test case for formatting strings

rchiodo:


heejaechang (approved):

testing/single/src/formatOnType.py
```diff
@@ -20,4 +20,22 @@
...
+w = ''''''
+
+# Make sure it doesn't add the `f` if it's already there
+a = f""
+b = f''
+c = f"""""""
+d = f''''''
+
+# Make sure it doesn't add the `f` if it the start is a r or b string.
+e = r""
```
debonte, line 40: `rf` is supported by Python btw.

debonte (approved):

testing/single/src/formatOnType.py
rchiodo: Oh well i guess we should support it too. It's not currently supported by our algorithm though.

rchiodo: Added an issue https://github.com/microsoft/pylance-release/issues/5703

PR #5846, opened 2024-05-03
[WIP] added more troubleshoot for memory

heejaechang:


bschnurr (approved):
love it. nice work

TROUBLESHOOTING.md
```diff
@@ -101,6 +101,16 @@ To specify your own Node.js executable, set this setting in your User settings.j
 "python.analysis.nodeExecutable": "<path to node.js exe>"
 ```
 
+For those using vscode-server remotely, you can increase the memory limit by setting the `NODE_OPTIONS` environment variable in your shell configuration.
+
+On Linux or Mac, add `export NODE_OPTIONS="--max-old-space-size=8192"` to either your `.xxx_profile` or `.xxxrc` file. On Windows, add `set NODE_OPTIONS=--max-old-space-size=8192` to your system environment variables.
```
StellaHuang95, line 106: "add `set NODE_OPTIONS=--max-old-space-size=8192` to your system environment variables." Maybe remove `set` to avoid confusion?

heejaechang: `set` is part of command. not part of english phrase.

rchiodo (approved):

TROUBLESHOOTING.md
StellaHuang95: ok missed the first part that's in shell config.

StellaHuang95 (approved):

TROUBLESHOOTING.md
heejaechang: I am open to any other suggestion. should I say add `set ...` to your batch file to add it to your system environment variable?

StellaHuang95: When I read the node.js link you put in there, my first thought was to open `System Properties` and copy the `Environment Variable` there, so yes I think what you have above is more clear to me.

heejaechang: how about now?

PR #5949, opened 2024-05-31
Add comment in formatOnType to indicate autoIndent is set to 'none'

rchiodo:
Not sure if this was on purpose, but it seems like we'd want to test with the defaults?


rchiodo:
I'll add some comments into formatOnType.py to change this setting on and off and test with both.

debonte (approved):

StellaHuang95 (approved):

PR #5989, opened 2024-06-10
Adding doc string tests to manual pylance-release tests

rchiodo:
Adding a bunch of examples for things that should stay working in docstring formatting. Test guidance isn't very prescriptive though. It's basically a judgement call.


testing/single/src/notebook.ipynb
```diff
@@ -602,9 +602,7 @@
    "outputs": [],
    "source": [
     "# change notebook kernel to different one and confirm notebook doesn't crash \n",
-    "# and notebook react accordingly. (some imports are no longer resolved and semantic/inlay refreshs as expected)\n",
```
rchiodo, line 605: Hmm, maybe this is still broken. I'll undo this change.

testing/single/src/docstrings.py
```diff
@@ -0,0 +1,1295 @@
+# This file is for testing different types of docstrings and verifying they display
+# correctly.
+# Try it with `python.analysis.supportRestructuredText` both true and false. 
+import numpy as np
```
StellaHuang95, line 4: Should `numpy` also be added to requirements.txt?

rchiodo: Ah good idea. I'll do that.

testing/single/src/docstrings.py
```diff
@@ -0,0 +1,1295 @@
...
+
+# Hover over zeroes and ensure the markdown looks correct.
+z = np.zeros([1, 2, 3])
+
+# Hover over take and ensure the markdown looks correct.
```
StellaHuang95, line 9: Can you be more specific about what looks correct or rendered correctly means here so that we know what to look for?

rchiodo: I can't. That's the whole problem. It's subjective. I'll say that.

debonte (approved):

StellaHuang95 (approved):

PR #6069, opened 2024-06-28
Simplify and update bug-report.md

dandv:
* The instructions to find the version of the extension that had the user toggle the Output then select "Python language server" then look at the  "Pylance language server" were unnecessarily complicated compared to simply looking at the extension version next to the its name. Also, it wasn't clear what "Language Server version" was (vs. the much clearer "Pylance version".

* Peek is an awesome GIF recorder but has been deprecated for a few years now. LiceCAP doesn't support Linux. VokoscreenNG is open-source and cross-platform, so I placed it first in the list.


debonte (approved):

luabud (approved):
Apologies for the delay, but thanks for your contribution!

PR #6185, opened 2024-07-25
new sklearn notebook tests

judej:
Added sklearn tests


testing/scikitlearn/test2.py
```diff
@@ -0,0 +1,73 @@
+from sklearn.model_selection import KFold
```
rchiodo, line 1: Do you want to add comments outlining what to test? Would it be better to generate all of this code as you tested?

rchiodo: This kinda feels like what you should end up with.

PR #6813, opened 2025-01-03
added documents for `include/exclude/ingore/extraPaths`

heejaechang:
added a few more docs for pylance settings.

### DIFF
--- a/docs/settings/python_analysis_ignore.md
+++ b/docs/settings/python_analysis_ignore.md
@@ -1,0 +1,136 @@
+# Understanding `python.analysis.ignore` in Pylance
+
+[Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) is a fast, feature-rich language support extension for Python in Visual Studio Code, powered by the Pyright static type checker. It provides advanced type checking, auto-completions, code navigation, and other language features to enhance your Python development experience.
+
+In large projects or specific development environments, you might encounter situations where certain files or directories generate unwanted diagnostics (errors and warnings) from Pylance. This can occur with third-party libraries, generated files, or code that intentionally deviates from standard practices. To manage this, Pylance provides the `python.analysis.ignore` setting, allowing you to suppress diagnostics for specified paths.
+
+This guide explains what `python.analysis.ignore` does, how to configure it, and how it can help you tailor Pylance's behavior to your project's needs.
+
+## What is `python.analysis.ignore`?
+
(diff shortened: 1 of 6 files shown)


docs/settings/python_analysis_ignore.md
```diff
@@ -0,0 +1,136 @@
...
+{
+    "python.analysis.ignore": ["src/generated/**"]
+}
+```
+
+This way, Pylance won't show errors or warnings for the generated code but will still provide IntelliSense when you import from it.
+
+### Ignoring Files in Virtual Environments
+
+Sometimes, Pylance may display diagnostics for files in your virtual environment. You can suppress these by ignoring the virtual environment directories:
```
StellaHuang95, line 95: Does this need to be explicitly set? If `exclude` is not specified by user, venv is ignored by default, right?

heejaechang: ya, below, it talks about relationship between `exclude/ignore` but welcome any suggestions if you have better wording.

docs/settings/python_analysis_include.md
```diff
@@ -0,0 +1,174 @@
+# Understanding `python.analysis.include` in Pylance
+
+[Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) is a fast, feature-rich language support extension for Python in Visual Studio Code, powered by the Pyright static type checker. It provides advanced type checking, IntelliSense, code navigation, and other language features to enhance your Python development experience.
```
StellaHuang95, line 3: The first paragraph feels redundant to include in every document, but I'll leave it up to you to decide whether to keep it.

StellaHuang95 (approved):

docs/settings/python_analysis_ignore.md
StellaHuang95: Maybe add a clarification saying that if `exclude` is not set, the same default directories such - `**/node_modules`, `**/__pycache__`, `xxx`, directories are ignored by default.

heejaechang: but it is not `ignored`, it is `excluded` in that case. 2 are different concept.

heejaechang: I guess I can remove some of these examples since it is more confusing then clarifying things.

StellaHuang95: I thought the `venv` is ignored by default, because I don't see any diagnostics in the error list generated for the files in `venv` directory.

docs/settings/python_analysis_include.md
heejaechang: that's current pattern but we could decide what pattern we want later on if we want to change. that said, I think since all documents are independent, not sure we should assume what order people will see these docs.

PR #6817, opened 2025-01-04
added a few more wikis

heejaechang:


README.md
```diff
@@ -216,7 +216,13 @@ Pylance provides users with the ability to customize their Python language suppo
     - Performance Consideration:
         - Adjusting this setting will cause Pylance to allocate more resources for indexing third-party libraries.
 
-- `python.analysis.regenerateStdLibIndices`
+- [`python.analysis.persistAllIndices`](docs/settings/python_analysis_persistAllIndices.md)
+    - Used to specify whether indices for all third-party libraries should be persisted to disk.
+    - Default value: `true`
+    - Performance Consideration:
+        - Enabling this setting can improve performance at load by reducing the need to re-index third-party libraries.
```
luabud, line 223:
```suggestion
        - Enabling this setting can improve startup performance by reducing the need to re-index third-party libraries.
```

docs/settings/python_analysis_persistAllIndices.md
```diff
@@ -0,0 +1,143 @@
...
+
+One of the ways Pylance provides these features is through **indexing** your project's files and installed libraries. Indexing allows Pylance to analyze the codebase and offer features like auto-import suggestions and workspace symbol search more efficiently.
+
+The `python.analysis.persistAllIndices` setting in Pylance controls whether the indices for third-party libraries are saved to disk and reused across sessions. This guide explains what indexing is in Pylance, how the `python.analysis.persistAllIndices` setting affects it, and how you can configure it to optimize your development workflow.
+
+## What Is Indexing in Pylance?
+
+Indexing is the process by which Pylance scans your project's files and installed third-party libraries to build a database (indices) of symbols such as classes, functions, and variables. This enables features like:
```
StellaHuang95, line 11: Just curious, the concept of database here is different than the concept of database in web services right? What is the form of this database? (For example, for web services it could be a relational database like SQL).

docs/settings/python_analysis_persistAllIndices.md
```diff
@@ -0,0 +1,143 @@
...
+### Ideal Scenarios for Enabling
+
+- **Stable Dependencies**: If your project's dependencies don't change frequently, persisting indices ensures that Pylance always has up-to-date information without needing to re-index.
+- **Large Projects**: In projects with a lot of third-party dependencies, indexing can take considerable time. Persisting indices mitigates this overhead.
+- **Continuous Development**: For developers who open and close their projects frequently, persisting indices enhances productivity by reducing wait times.
+
+### Potential Considerations
+
+- **Disk Space Usage**: Persisted indices consume disk space. The amount depends on the size and number of third-party libraries.
+- **Outdated Indices**: If you update or add new packages, the persisted indices might become outdated. In such cases, you may need to clear and regenerate the indices.
```
StellaHuang95, line 60: Does this mean if this setting is set to `false`, and user installs a new third party library in their environment, it would cause some features like `auto import` to NOT work for this newly installed package?

docs/settings/python_analysis_persistAllIndices.md
```diff
@@ -0,0 +1,143 @@
...
+    },
+    {
+      "name": "pandas",
+      "depth": 2,
+      "includeAllSymbols": false
+    }
+  ]
+  ```
+
+- **Note**: Increasing the index depth can improve auto-import suggestions but may impact indexing time and resource usage. When this setting is changed, indices need to be regenerated by clearing persisted indices.
```
luabud, line 106:
```suggestion
### [`python.analysis.indexing`](python_analysis_indexing.md)

- Determines whether Pylance should index user files and installed third-party libraries.
- **Note**: If `indexing` is set to `false`, `persistAllIndices` has no effect because indices are not created.

### [`python.analysis.packageIndexDepths`](python_analysis_packageIndexDepths.md)

- Configures the depth of indexing for specific packages.
- **Note**: Increasing the index depth can improve auto-import suggestions but may impact indexing time and resource usage. When this setting is changed, indices need to be regenerated by clearing persisted indices. 
```

luabud: this suggestion is to mainly keep the doc a bit shorter and keep consistency with the "related settings" section format from the autoImportCompletion page

docs/settings/python_analysis_regenerateStdLibIndices.md
```diff
@@ -0,0 +1,99 @@
...
+## Limitations and Considerations
+
+- **Workspace-Specific**: The indices are regenerated per workspace. If you work with multiple projects, you need to ensure the setting is enabled in each workspace.
+- **Initial Delay**: There may be a delay when opening a workspace as Pylance regenerates the indices.
+- **Persistence of Indices**: With [`persistAllIndices`](python_analysis_persistAllIndices.md) settings, this can be persisted to disk and happen only once. Otherwise, it will happen every time for a new session.
+
+## Related Settings
+
+- **[`python.analysis.indexing`](python_analysis_indexing.md)**: Used to specify whether Pylance should index installed third party libraries and user files to improve features such as auto-import, add import, workspace symbols, etc.
+- **[`python.analysis.persistAllIndices`](python_analysis_persistAllIndices.md)**: Determines if third-party library indices are preserved across VS Code sessions.
```
luabud, line 91:
```suggestion
- [`python.analysis.indexing`](python_analysis_indexing.md): 
   - Used to specify whether Pylance should index installed third party libraries and user files to improve features such as auto-import, add import, workspace symbols, etc.
- [`python.analysis.persistAllIndices`]
(python_analysis_persistAllIndices.md): 
   - Determines if third-party library indices are preserved across VS Code sessions.
```

docs/settings/python_analysis_persistAllIndices.md
luabud: that's a good point that the term could be confusing to a few users. Maybe a "cache" instead of "database"?

luabud (approved):

docs/settings/python_analysis_autoImportCompletions.md
```diff
@@ -0,0 +1,128 @@
...
+Selecting the suggestion automatically adds the import statement at the top of your file:
+
+```python
+from mathlib.shapes import Circle
+
+def calculate_area():
+    circle = Circle(radius=5)
+```
+
+This feature streamlines the coding process by reducing the need to manually write import statements.
```
gramster, line 35: This suggests a feature we should add if we don't have it already, which is to change imports like:

    from M import A, B
    ...
    A()
    ...
    B()

to:

    import M
    ...
    M.A()
    ...
    M.B()

as the latter is better practice.

erictraut: Why is the latter considered better practice? I don't recall seeing that guideline before. I find the former to be less cluttered and more readable, so I write most of my code that way.

gramster: It would be interesting to get Brett or Steve to weigh in here; I may just be reflecting a bias I picked up from when I used to work at Google:

https://google.github.io/styleguide/pyguide.html#22-imports

docs/settings/python_analysis_persistAllIndices.md
heejaechang: it might, that depends on file watcher and other stuff. if everything works as expected, indices will be updated, and persisted indices will be also updated. but when that fails (for whatever reasons), if this is `false`, in next session, the new package will be discovered since we will always re-index, if this is `true`, we will use `persisted indices` so it won't get fixed until user manually clear the persisted cache. it is typical issue of using cache. cache invalidation failure case.

heejaechang: the section is there so when feature seems not not working as expected, user has a something to try regardless of whether that's the issue or not. it is like when something not working, people simply tell users to reboot your OS. and many times, that fixes issues regardless of the root cause.

docs/settings/python_analysis_persistAllIndices.md
heejaechang: just to let you know, docs' style in the docs/settings folder is not currently 100% consistent. most of this content is same as what AI generated from various sources as all other docs in the folder are.

docs/settings/python_analysis_autoImportCompletions.md
heejaechang: I think we had similar feature requests before, but I couldn't find one so I added new one - https://github.com/microsoft/pylance-release/issues/6818

but anyway, I think Brett also mentioned he wants this feature as well. 

currently, pylance does do this though

```python
import M # import statement already exists.

A() # auto import here
```

we will change `A` to `M.A` instead of adding `from M import A`

docs/settings/python_analysis_persistAllIndices.md
heejaechang: I simply removed `database` and just say `build indices`

PR #7233, opened 2025-05-16
Add Type Server Protocol first revision

rchiodo:
First revision of the Type Server Protocol being used for semantic tokens

### DIFF
--- a/docs/tsp/typeServerProtocol.ts
+++ b/docs/tsp/typeServerProtocol.ts
@@ -1,0 +1,383 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ * All rights reserved.
+ *
+ * typeServerProtocol.ts
+ *
+ * Defines the interfaces and types for the type server protocol. A Type Server is a module that provides type information
+ * for code, such as type definitions, member information, and diagnostics.
+ *
+ * It's Python specific at the moment, but may be made generic in the future.
+ *
+ * This protocol is used to communicate between the type server and the client (e.g., a language server or an IDE).
+ *
+ * All the types in this file should be JSON serializable, as they are sent over the wire.
+ * The protocol is designed to be extensible, allowing for future additions of new requests and notifications.
+ */
+import {
+    CancellationToken,
+    Diagnostic,
+    Disposable,
+    FileEvent,
+    NotificationHandler,
+    RequestHandler,
+    WorkspaceFolder,
+} from 'vscode-languageserver';
+
+export interface RequestSender {
+    sendRequest<R>(method: string, params: any, token?: CancellationToken): Promise<R>;
+}
+
+export interface NotificationSender {
+    sendNotification: (method: string, params?: any) => void;
+}
+
+export interface RequestReceiver {
+    onRequest<P, R, E>(method: string, handler: RequestHandler<P, R, E>): Disposable;
+}
+
+export interface NotificationReceiver {
+    onNotification<P>(method: string, handler: NotificationHandler<P>): Disposable;
+}
+
+export namespace TypeServerProtocol {
+    // Represents a node in an AST (Abstract Syntax Tree) or similar structure.
+    export interface Node {
+        // URI of the source file containing this node.
+        uri: string;
+        // Hash of the content of the source file at the time the AST was created.
+        contentHash: number;
+        // The start byte position (zero-based) of the node in the source file.
+        // Note this is per byte and not per character.
+        start: number;
+        // The length of the node (in number of bytes) in the source file.
+        length: number;
+    }
+
+    // Represents a category of a type, such as class, function, variable, etc.
+    export const enum TypeCategory {
+        Class = 0,
+        Function = 1,
+        Overloaded = 2,
+        Module = 3,
+        Variable = 4,
+        Parameter = 5,
+        Property = 6,
+        TypeAlias = 7,
+        TypeVariable = 8,
+        Generic = 9,
+        Unknown = 10,
+        Any = 11,
+        None = 12,
+        Union = 13,
+    }
+
+    // Flags that describe the characteristics of a type.
+    // These flags can be combined using bitwise operations.
+    export const enum TypeFlags {
+        None = 0,
+        Instantiable = 1 << 0, // Indicates if the type can be instantiated.
+        Instance = 1 << 1, // Indicates if the type represents an instance (as opposed to a class or type itself).
+        Callable = 1 << 2, // Indicates if an instance of the type can be called like a function. (It has a `__call__` method).
+        BuiltIn = 1 << 3, // Indicates if the type is a built-in type (like `int`, `str`, etc.).
+        Literal = 1 << 4, // Indicates if the instance is a literal (like `42`, `"hello"`, etc.).
+        Interface = 1 << 5, // Indicates if the type is an interface (a type that defines a set of methods and properties). In Python this would be a Protocol.
+        Generic = 1 << 6, // Indicates if the type is a generic type (a type that can be parameterized with other types).
+    }
+
+    // Flags that describe the characteristics of a function or method.
+    // These flags can be combined using bitwise operations.
+    export const enum FunctionFlags {
+        None = 0,
+        Async = 1 << 0, // Indicates if the function is asynchronous.
+        Generator = 1 << 1, // Indicates if the function is a generator (can yield values).
+        Abstract = 1 << 2, // Indicates if the function is abstract (must be implemented in a subclass).
+        Static = 1 << 3, // Indicates if the function is static (belongs to the class rather than an instance).
+    }
+
+    // Flags that describe the characteristics of a class.
+    // These flags can be combined using bitwise operations.
+    export const enum ClassFlags {
+        None = 0,
+        Enum = 1 << 0, // Indicates if the class is an enum (a special kind of class that defines a set of named values).
+    }
+
+    export interface Type {
+        // Unique identifier for the type definition within the session.
+        handle: string | number;
+
+        // Essential classification of the symbol.
+        category: TypeCategory;
+
+        // Flags describing the type.
+        flags: TypeFlags;
+
+        // Name of the module the type comes from
+        moduleName: string | undefined;
+
+        // Name of the type.
+        name: string;
+
+        // Flags specific to the category
+        categoryFlags: number;
+    }
+
+    export interface Member {
+        // The name of the member.
+        name: string;
+
+        // The type of the member.
+        type: Type;
+
+        // The type the member came from
+        classType: Type | undefined;
+    }
+
+    // Flags that are used for searching for members of a type.
+    export const enum MemberAccessFlags {
+        None = 0,
+        SkipInstanceMembers = 1 << 0, // Skip instance members when searching for members of a type.
+        SkipObjectBaseClass = 1 << 1, // Skip members from the `object` base class when searching for members of a type.
+    }
+
+    // Represents the category of a declaration in the type system.
+    // This is used to classify declarations such as variables, functions, classes, etc.
+    export const enum DeclarationCategory {
+        Intrinsic = 0, // An intrinsic is a built-in type or function that is part of the language itself, such as `int`, `str`, `len`, etc.
+        Variable = 1, // A variable is a named storage location that can hold a value.
+        Param = 2, // A parameter is a variable that is passed to a function or method.
+        TypeParam = 3, // A type parameter is a placeholder for a type that can be specified when the type is used, such as in generics.
+        TypeAlias = 4, // A type alias is a name that refers to another type, such as `List[int]` or `Dict[str, int]`.
+        Function = 5, // A function is a named block of code that can be called with arguments and returns a value.
+        Class = 6, // A class is a blueprint for creating objects that encapsulates data and behavior.
+        SpecialBuiltInClass = 7, // Special built-in types like 'Tuple', 'Generic',  'Protocol', 'Callable', 'Type', etc
+        Alias = 8, // An alias declaration, which is a reference to another declaration.
+    }
+
+    // Represents a symbol declaration in the type system.
+    // A declaration is a specific instance of a symbol in the source code, such as a variable, function, or class.
+    export interface Declaration {
+        // Unique identifier for the declaration within the session.
+        handle: string | number;
+
+        // Category of this symbol (function, variable, etc.).
+        category: DeclarationCategory;
+
+        // Parse node associated with the declaration
+        node?: Node;
+
+        // The dot-separated import name for the file that
+        // contains the declaration (may not be definitive
+        // because a source file can be accessed via different
+        // import names in some cases).
+        moduleName: string;
+    }
+
+    // Synthesized type information for a declaration that is not directly
+    // represented in the source code, but is derived from the declaration. Other languages
+    // may refer to this as an anonymous type.
+    export interface SynthesizedTypeInfo {
+        type: Type;
+
+        // An optional node that is not used by the type evaluator
+        // but can be used by language services to provide additional
+        // functionality (such as go-to-definition).
+        node?: Node;
+    }
+
+    // Declaration information for a symbol, which includes
+    // a list of declarations and potentially synthesized types for those declarations.
+    export interface SymbolDeclInfo {
+        decls: Declaration[];
+        synthesizedTypes: SynthesizedTypeInfo[];
+    }
+
+    // Describes an import
+    export interface ImportedModuleDescriptor {
+        leadingDots: number;
+        nameParts: string[];
+    }
+
+    // Options for resolving an alias declaration.
+    export interface ResolveAliasOptions {
+        resolveLocalNames?: boolean; // Whether to resolve local names in the alias declaration.
+        allowExternallyHiddenAccess?: boolean; // Whether to allow access to members that are hidden by external modules.
+        skipFileNeededCheck?: boolean; // Whether to skip checking if the file is needed for the alias resolution.
+    }
+
+    // Describes a file that's been opened in the language server which the type server needs to be aware of.
+    export interface TextDocumentOpenParams {
+        uri: string;
+        text: string;
+        // The version of the file, which is used to track changes to the file. It will be incremented each time the file is modified.
+        version: number;
+        // Optional: If the file is part of a chain of files, this can be used to indicate the URI of the
+        // file that is automatically 'imported' by this file. This is useful for cases like Jupyter notebooks
+        chainedFileUri?: string;
+    }
+
+    // Describes a file that's been closed in the language server which the type server needs to be aware of.
+    export interface TextDocumentCloseParams {
+        uri: string;
+    }
+
+    // Parameters for resolving an import
+    export interface ResolveImportParams {
+        // The URI of the source file where the import is referenced.
+        sourceUri: string;
+        // The descriptor of the imported module.
+        moduleDescriptor: ImportedModuleDescriptor;
+        snapshot: number;
+    }
+
+    // Represents settings that can be sent to the type server.
+
+    export type Settings = {
+        [key: string]: any;
+    };
+
+    // Requests and notifications for the type server protocol.
+    export enum Requests {
+        // First request to initialize the type server.
+        Initialize = 'typeServer/initialize',
+        // Request from client to shut down the type server.
+        Shutdown = 'typeServer/shutdown',
+        // Request from client to get the current snapshot of the type server.
+        // A snapshot is a point-in-time representation of the type server's state, including all loaded files and their types.
+        // A type server should change its snapshot whenever any type it might have returned is no longer valid. Meaning types are
+        // only usable for the snapshot they were returned with.
+        GetSnapshot = 'typeServer/getSnapshot',
+        // Request to get diagnostics for a specific file.
+        GetDiagnostics = 'typeServer/getDiagnostics',
+        // Request to get the version of diagnostics for a specific file.
+        GetDiagnosticsVersion = 'typeServer/getDiagnosticsVersion',
+        // Request to get the type information for a specific node.
+        GetType = 'typeServer/getType',
+        // Request to get the union types for a specific type.
+        GetUnionTypes = 'typeServer/getUnionTypes',
+        // Request to find a member of a type.
+        GetTypeMember = 'typeServer/getTypeMember',
+        // Request to get all overloads of a function or method.
+        GetOverloads = 'typeServer/getOverloads',
+        // Request to get the effective meta class of a type.
+        GetEffectiveMetaClass = 'typeServer/getEffectiveMetaClass',
+        // Request to get the type of a declaration.
+        GetTypeOfDeclaration = 'typeServer/getTypeOfDeclaration',
+        // Request to get symbol declaration information for a node.
+        GetSymbolDeclarationInfo = 'typeServer/getSymbolDeclarationInfo',
+        // Request to get the bound magic method for a type.
+        GetBoundMagicMethod = 'typeServer/getBoundMagicMethod',
+        // Request to resolve an alias declaration.
+        // This request is sent by the client to resolve an alias declaration, which may involve looking up the
+        // declaration in the type server's cache or performing additional resolution logic.
+        ResolveAliasDeclaration = 'typeServer/resolveAliasDeclaration',
+        // Request to resolve an import.
+        ResolveImport = 'typeServer/resolveImport',
+    }
+
+    export enum Notifications {
+        // Notification sent by the server to indicate that the type server has been initialized.
+        Initialized = 'typeServer/initialized',
+        // Notification sent by the server to indicate that the type server has shut down.
+        Shutdown = 'typeServer/shutdown',
+        // Notification sent by the client to indicate that a text document has been opened.
+        TextDocumentOpen = 'textDocument/open',
+        // Notification sent by the client to indicate that a text document has been closed.
+        TextDocumentClose = 'textDocument/close',
+        // Notification sent by the client to indicate that settings have changed.
+        SettingsChange = 'typeServer/settingsChange',
+        // Notification sent by the client to indicate that files have changed.
+        FilesChanged = 'typeServer/filesChanged',
+        // Notification sent by the client to indicate that the type server's cache should be flushed (and snapshots reset)
+        FlushCache = 'typeServer/flushCache',
+    }
+
+    export interface Params {
+        [Requests.Initialize]: { workspace: WorkspaceFolder; initialSettings: string };
+        [Requests.Shutdown]: void;
+        [Requests.GetSnapshot]: void;
+        [Requests.GetDiagnostics]: { uri: string; snapshot: number };
+        [Requests.GetDiagnosticsVersion]: { uri: string };
+        [Requests.GetType]: { node: Node; snapshot: number };
+        [Requests.GetUnionTypes]: { type: Type; snapshot: number };
+        [Requests.GetTypeMember]: {
+            type: Type;
+            memberName: string;
+            accessFlags: MemberAccessFlags;
+            snapshot: number;
+        };
+        [Requests.GetOverloads]: { type: Type; snapshot: number };
+        [Requests.GetEffectiveMetaClass]: { type: Type; snapshot: number };
+        [Requests.GetTypeOfDeclaration]: { decl: Declaration; snapshot: number };
+        [Requests.GetSymbolDeclarationInfo]: {
+            node: Node;
+            skipUnreachableCode: boolean;
+            snapshot: number;
+        };
+        [Requests.GetBoundMagicMethod]: { type: Type; name: string; snapshot: number };
+        [Requests.ResolveAliasDeclaration]: {
+            decl: Declaration;
+            options: ResolveAliasOptions;
+            snapshot: number;
+        };
+        [Requests.ResolveImport]: ResolveImportParams;
+        [Notifications.Initialized]: void;
+        [Notifications.Shutdown]: void;
+        [Notifications.FlushCache]: void;
+        [Notifications.TextDocumentOpen]: TextDocumentOpenParams;
+        [Notifications.TextDocumentClose]: TextDocumentCloseParams;
+        [Notifications.SettingsChange]: Settings;
+        [Notifications.FilesChanged]: { changes: FileEvent[] };
+    }
+
+    export interface Response {
+        [Requests.Initialize]: { configRoots: string[] };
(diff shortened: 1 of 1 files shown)


rchiodo:
This is submitting this same file from pyrx. I'm going to point the discussion item at it.

debonte (approved):

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        None = 0,
+        Enum = 1 << 0, // Indicates if the class is an enum (a special kind of class that defines a set of named values).
+    }
+
+    export interface Type {
+        // Unique identifier for the type definition within the session.
+        handle: string | number;
+
+        // Essential classification of the symbol.
+        category: TypeCategory;
```
erictraut, line 110: The comment says "classification of the symbol". What "symbol" is this referring to? If this interface is describing a "type", then a "symbol" should not be involved. Generally speaking, symbols have types, but types don't have symbols.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+
+        // Flags describing the type.
+        flags: TypeFlags;
+
+        // Name of the module the type comes from
+        moduleName: string | undefined;
+
+        // Name of the type.
+        name: string;
```
erictraut, line 119: I presume this is the "short" type name, not its fully-qualified name.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+
+        // Flags specific to the category
+        categoryFlags: number;
+    }
+
+    export interface Member {
```
erictraut, line 125: In python terminology, "Attribute" would be a better name for this concept. I'm not sure if the TSP is intended to be Python-specific.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        // The name of the member.
+        name: string;
+
+        // The type of the member.
+        type: Type;
+
+        // The type the member came from
+        classType: Type | undefined;
```
erictraut, line 133: When you say "came from", are you referring to where it is declared? Or what it is bound to? 

For example:
```python
class A:
    def m(self) -> None: ...

class B(A): ...

b = B()
b.m
```

If we're describing the attribute `b.m`, would `classType` be `A` or `B` in this case?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    // Flags that are used for searching for members of a type.
+    export const enum MemberAccessFlags {
```
erictraut, line 137: When you say "members of a type", is this limited to classes? What about modules? Does it apply to unions and intersections? Based on the provided flags, I'm guessing that it's limited to just classes?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        None = 0,
+        SkipInstanceMembers = 1 << 0, // Skip instance members when searching for members of a type.
+        SkipObjectBaseClass = 1 << 1, // Skip members from the `object` base class when searching for members of a type.
```
erictraut, line 140: We can probably omit this flag to simplify the interface. Pyright uses this flag internally, but other type checkers will likely not. A caller doesn't really need this functionality if the member information includes the information about where the attribute was declared. If it's declared in the `object` class, the caller can ignore it if desired.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    // Represents the category of a declaration in the type system.
+    // This is used to classify declarations such as variables, functions, classes, etc.
+    export const enum DeclarationCategory {
+        Intrinsic = 0, // An intrinsic is a built-in type or function that is part of the language itself, such as `int`, `str`, `len`, etc.
```
erictraut, line 146: This comment is a misleading explanation of what an "intrinsic" declaration is. After all, the symbols `int`, `str`, and `len` all have explicit type declarations in `builtins.pyi`. An intrinsic declaration is one that has no explicit declaration within the code. An example is the `__class__` symbol within a class body. Pyright has also historically used intrinsic declarations for certain dundered symbols that are omitted from the typeshed stubs, but those stubs have become more complete over time, so the need for intrinsic declarations has decreased.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Variable = 1, // A variable is a named storage location that can hold a value.
+        Param = 2, // A parameter is a variable that is passed to a function or method.
+        TypeParam = 3, // A type parameter is a placeholder for a type that can be specified when the type is used, such as in generics.
+        TypeAlias = 4, // A type alias is a name that refers to another type, such as `List[int]` or `Dict[str, int]`.
+        Function = 5, // A function is a named block of code that can be called with arguments and returns a value.
+        Class = 6, // A class is a blueprint for creating objects that encapsulates data and behavior.
+        SpecialBuiltInClass = 7, // Special built-in types like 'Tuple', 'Generic',  'Protocol', 'Callable', 'Type', etc
+        Alias = 8, // An alias declaration, which is a reference to another declaration.
```
erictraut, line 154: I recommend renaming "Alias" to "Import" for clarity. I somewhat regret the "alias" terminology I adopted in pyright because it is easy to confuse it with a "type alias".

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        SkipObjectBaseClass = 1 << 1, // Skip members from the `object` base class when searching for members of a type.
+    }
+
+    // Represents the category of a declaration in the type system.
+    // This is used to classify declarations such as variables, functions, classes, etc.
+    export const enum DeclarationCategory {
+        Intrinsic = 0, // An intrinsic is a built-in type or function that is part of the language itself, such as `int`, `str`, `len`, etc.
+        Variable = 1, // A variable is a named storage location that can hold a value.
+        Param = 2, // A parameter is a variable that is passed to a function or method.
+        TypeParam = 3, // A type parameter is a placeholder for a type that can be specified when the type is used, such as in generics.
```
erictraut, line 149: This is specifically for PEP 695 type parameters. It doesn't apply to older type parameters, which use variable declarations.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        TypeAlias = 4, // A type alias is a name that refers to another type, such as `List[int]` or `Dict[str, int]`.
```
erictraut, line 150: This is specifically for PEP 695 type aliases. It doesn't apply to older type aliases, which use variable declarations.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Function = 5, // A function is a named block of code that can be called with arguments and returns a value.
```
erictraut, line 151: This specifically refers to declarations that use the `def` statement.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Class = 6, // A class is a blueprint for creating objects that encapsulates data and behavior.
```
erictraut, line 152: This specifically refers to declarations that use the `class` statement.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        SpecialBuiltInClass = 7, // Special built-in types like 'Tuple', 'Generic',  'Protocol', 'Callable', 'Type', etc
```
erictraut, line 153: The "SpecialBuiltInClass" declaration type is very specific to pyright. I recommend omitting it from this list. It could be modeled as a class for purposes of this interface.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        // contains the declaration (may not be definitive
+        // because a source file can be accessed via different
+        // import names in some cases).
+        moduleName: string;
+    }
+
+    // Synthesized type information for a declaration that is not directly
+    // represented in the source code, but is derived from the declaration. Other languages
+    // may refer to this as an anonymous type.
+    export interface SynthesizedTypeInfo {
```
erictraut, line 179: This is pretty specific to pyright, but I don't see how to eliminate it from the interface.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    // Describes an import
+    export interface ImportedModuleDescriptor {
+        leadingDots: number;
+        nameParts: string[];
+    }
+
+    // Options for resolving an alias declaration.
+    export interface ResolveAliasOptions {
```
erictraut, line 202: As mentioned above, I recommend renaming this from "Alias" to "Import".

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        resolveLocalNames?: boolean; // Whether to resolve local names in the alias declaration.
```
erictraut, line 203: These options are pretty specific to pyright's internal implementation. Can the language server reduce or eliminate its dependence on them? If not, could we provide a better description and/or abstraction?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        allowExternallyHiddenAccess?: boolean; // Whether to allow access to members that are hidden by external modules.
+        skipFileNeededCheck?: boolean; // Whether to skip checking if the file is needed for the alias resolution.
+    }
+
+    // Describes a file that's been opened in the language server which the type server needs to be aware of.
+    export interface TextDocumentOpenParams {
```
erictraut, line 209: What is the file ownership model proposed by the TSP? Does the type server own the contents of closed files? Who is responsible for detecting on-disk changes? Is that the type server or the language server?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    export enum Requests {
+        // First request to initialize the type server.
+        Initialize = 'typeServer/initialize',
+        // Request from client to shut down the type server.
+        Shutdown = 'typeServer/shutdown',
+        // Request from client to get the current snapshot of the type server.
+        // A snapshot is a point-in-time representation of the type server's state, including all loaded files and their types.
+        // A type server should change its snapshot whenever any type it might have returned is no longer valid. Meaning types are
+        // only usable for the snapshot they were returned with.
+        GetSnapshot = 'typeServer/getSnapshot',
```
erictraut, line 249: I presume this returns some sort of an ID that uniquely identifies the current state? Or does it actively create a new snapshot object whose lifetime must be tracked and managed by the type server?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        // Request to get diagnostics for a specific file.
+        GetDiagnostics = 'typeServer/getDiagnostics',
+        // Request to get the version of diagnostics for a specific file.
+        GetDiagnosticsVersion = 'typeServer/getDiagnosticsVersion',
+        // Request to get the type information for a specific node.
+        GetType = 'typeServer/getType',
+        // Request to get the union types for a specific type.
+        GetUnionTypes = 'typeServer/getUnionTypes',
```
erictraut, line 257: I presume that `GetUnionTypes` is intended to return a collection of the subtypes that make up a union type?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    export type Settings = {
+        [key: string]: any;
+    };
+
+    // Requests and notifications for the type server protocol.
+    export enum Requests {
+        // First request to initialize the type server.
+        Initialize = 'typeServer/initialize',
+        // Request from client to shut down the type server.
+        Shutdown = 'typeServer/shutdown',
```
erictraut, line 244: Nit: "Shutdown" is an adjective, whereas "shut down" is a verb. It's analogous to `logout` and `log out`. I think you mean the latter, so it should be `ShutDown`.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        // Notification sent by the client to indicate that the type server's cache should be flushed (and snapshots reset)
+        FlushCache = 'typeServer/flushCache',
+    }
+
+    export interface Params {
+        [Requests.Initialize]: { workspace: WorkspaceFolder; initialSettings: string };
+        [Requests.Shutdown]: void;
+        [Requests.GetSnapshot]: void;
+        [Requests.GetDiagnostics]: { uri: string; snapshot: number };
+        [Requests.GetDiagnosticsVersion]: { uri: string };
```
erictraut, line 300: Wouldn't `GetDiagnosticsVersion` need to take a snapshot ID as well? If not, how will the caller know whether the version is still valid once the response arrives?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        GetTypeMember = 'typeServer/getTypeMember',
+        // Request to get all overloads of a function or method.
+        GetOverloads = 'typeServer/getOverloads',
+        // Request to get the effective meta class of a type.
+        GetEffectiveMetaClass = 'typeServer/getEffectiveMetaClass',
+        // Request to get the type of a declaration.
+        GetTypeOfDeclaration = 'typeServer/getTypeOfDeclaration',
+        // Request to get symbol declaration information for a node.
+        GetSymbolDeclarationInfo = 'typeServer/getSymbolDeclarationInfo',
+        // Request to get the bound magic method for a type.
```
erictraut, line 268: I think this should be changed from `GetBoundMagicMethod` to simply `GetBoundMethod`. There's nothing specific to "magic" methods about this call.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        // Request to get the type information for a specific node.
+        GetType = 'typeServer/getType',
+        // Request to get the union types for a specific type.
+        GetUnionTypes = 'typeServer/getUnionTypes',
+        // Request to find a member of a type.
+        GetTypeMember = 'typeServer/getTypeMember',
+        // Request to get all overloads of a function or method.
+        GetOverloads = 'typeServer/getOverloads',
+        // Request to get the effective meta class of a type.
+        GetEffectiveMetaClass = 'typeServer/getEffectiveMetaClass',
```
erictraut, line 263: Recommend renaming `GetEffectiveMetaClass` to `GetMetaclass`. (Note capitalization change as well.) The term "effective" is used internally to pyright to differentiate between declared and inherited metaclasses, but I don't think that language servers care about this distinction.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        [Requests.GetDiagnosticsVersion]: { uri: string };
+        [Requests.GetType]: { node: Node; snapshot: number };
+        [Requests.GetUnionTypes]: { type: Type; snapshot: number };
+        [Requests.GetTypeMember]: {
+            type: Type;
+            memberName: string;
+            accessFlags: MemberAccessFlags;
+            snapshot: number;
+        };
+        [Requests.GetOverloads]: { type: Type; snapshot: number };
```
erictraut, line 309: Should `GetOverloads` return the overload implementation signature as well as the overloads? Or should it filter out the overload implementation signature?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        // Notification sent by the client to indicate that a text document has been opened.
+        TextDocumentOpen = 'textDocument/open',
+        // Notification sent by the client to indicate that a text document has been closed.
+        TextDocumentClose = 'textDocument/close',
+        // Notification sent by the client to indicate that settings have changed.
+        SettingsChange = 'typeServer/settingsChange',
+        // Notification sent by the client to indicate that files have changed.
+        FilesChanged = 'typeServer/filesChanged',
+        // Notification sent by the client to indicate that the type server's cache should be flushed (and snapshots reset)
+        FlushCache = 'typeServer/flushCache',
```
erictraut, line 292: This is pretty specific to pyright. Shouldn't type cache management be an internal implementation of the type server? Why would the language server care about this?

erictraut (commented):
I tried to review the proposed interface through the lens of someone who doesn't know the internals of pyright. That's difficult to do, so it would be good to get feedback from other reviewers as well.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    onNotification<P>(method: string, handler: NotificationHandler<P>): Disposable;
+}
+
+export namespace TypeServerProtocol {
+    // Represents a node in an AST (Abstract Syntax Tree) or similar structure.
+    export interface Node {
+        // URI of the source file containing this node.
+        uri: string;
+        // Hash of the content of the source file at the time the AST was created.
+        contentHash: number;
```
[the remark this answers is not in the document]
erictraut: Ah, this represents the hash of the entire file. I thought you were proposing to hash the characters that comprise the AST node. Maybe rename to `fileContentHash` to make this clearer.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Any = 11,
+        None = 12,
+        Union = 13,
+    }
+
+    // Flags that describe the characteristics of a type.
+    // These flags can be combined using bitwise operations.
+    export const enum TypeFlags {
+        None = 0,
+        Instantiable = 1 << 0, // Indicates if the type can be instantiated.
```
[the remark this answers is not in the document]
rchiodo: We're using both the Instance and the Instantiable flag on the pylance side right now so we do need them. I'm not 100% sure of their meaning myself. Seems all class are instantiable? Unless maybe they're abstract? But then how is a type an instance? Is this a type class itself?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Instance = 1 << 1, // Indicates if the type represents an instance (as opposed to a class or type itself).
+        Callable = 1 << 2, // Indicates if an instance of the type can be called like a function. (It has a `__call__` method).
+        BuiltIn = 1 << 3, // Indicates if the type is a built-in type (like `int`, `str`, etc.).
```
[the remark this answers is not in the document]
rchiodo: We use this to make decisions about all sorts of stuff. We special case builtin classes in a lot of places. Maybe there's a less pyright specific way to do so, but for example this code is in the code action provider:

```
function isAddNode(evaluator: TypeEvaluator, node: ParseNode): node is BinaryOperationNode {
    if (node.nodeType !== ParseNodeType.BinaryOperation || node.d.operator !== OperatorType.Add) {
        return false;
    }

    const leftType = evaluator.getType(node.d.leftExpr);
    const rightType = evaluator.getType(node.d.rightExpr);
    if (!leftType || !rightType) {
        return false;
    }

    return (
        (isClassInstance(leftType) && ClassType.isBuiltIn(leftType, 'str')) || 
        (isClassInstance(rightType) && ClassType.isBuiltIn(rightType, 'str'))
    );
}
```

This code is checking to see if we're adding two strings (it's a code action that turns adding two strings into a format string). We could just check for 'str', but that's not as safe as making sure it's the builtin str class.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Literal = 1 << 4, // Indicates if the instance is a literal (like `42`, `"hello"`, etc.).
+        Interface = 1 << 5, // Indicates if the type is an interface (a type that defines a set of methods and properties). In Python this would be a Protocol.
```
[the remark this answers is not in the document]
rchiodo: Hmm, not sure. Up to the type server implementation?

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        Generic = 1 << 6, // Indicates if the type is a generic type (a type that can be parameterized with other types).
```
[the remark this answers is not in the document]
rchiodo: Not sure on this one either. To be determined. An open question I guess.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    // Flags that describe the characteristics of a function or method.
+    // These flags can be combined using bitwise operations.
+    export const enum FunctionFlags {
+        None = 0,
+        Async = 1 << 0, // Indicates if the function is asynchronous.
+        Generator = 1 << 1, // Indicates if the function is a generator (can yield values).
+        Abstract = 1 << 2, // Indicates if the function is abstract (must be implemented in a subclass).
+        Static = 1 << 3, // Indicates if the function is static (belongs to the class rather than an instance).
```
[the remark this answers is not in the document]
rchiodo: Thanks, I'll change it to say bound to a class.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+    // Flags that describe the characteristics of a class.
+    // These flags can be combined using bitwise operations.
+    export const enum ClassFlags {
+        None = 0,
+        Enum = 1 << 0, // Indicates if the class is an enum (a special kind of class that defines a set of named values).
+    }
+
+    export interface Type {
+        // Unique identifier for the type definition within the session.
+        handle: string | number;
```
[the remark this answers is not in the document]
rchiodo: Caller can't lookup a type, but the type server can use it to lookup internal extra information about a type. It's tied to the snapshot idea, when a snapshot changes, all type handles would be invalid. I should make that more explicit.

docs/tsp/typeServerProtocol.ts
rchiodo: Sorry bad comment. Should say classification of the Type

docs/tsp/typeServerProtocol.ts
rchiodo: Yes, I'll make this more explicit.

docs/tsp/typeServerProtocol.ts
rchiodo: Hmm, in the semantic token case, I believe this is 'A', but in working on other stuff we use 'B' too. Maybe we need two entries here which would give us A and B

docs/tsp/typeServerProtocol.ts
rchiodo: Limited to just classes. I'll change the comment.

docs/tsp/typeServerProtocol.ts
rchiodo: Okay will do.

docs/tsp/typeServerProtocol.ts
rchiodo: I'll fix the comment. Maybe we don't need this anymore? It is used by Pylance code but I think these could just be represented by the other categories?

docs/tsp/typeServerProtocol.ts
rchiodo: Thanks, I'll change the comment.

docs/tsp/typeServerProtocol.ts
rchiodo: I'll change the comment to mention that. Or are you suggesting we can drop this as it's not used much?

docs/tsp/typeServerProtocol.ts
rchiodo: Thanks. I'll change the comment.

docs/tsp/typeServerProtocol.ts
rchiodo: I mean that's how it works in Python, but maybe this could be more general? I guess we're Python for right now. I'll make it more specific.

docs/tsp/typeServerProtocol.ts
rchiodo: Yeah every spot this is used, we have it paired with just 'class', so we dont' treat it any differently. I'll remove it.

docs/tsp/typeServerProtocol.ts
rchiodo: Thanks, good idea. Will do.

docs/tsp/typeServerProtocol.ts
rchiodo: I'll look into that. I'll put a comment here to try and reduce them later.

docs/tsp/typeServerProtocol.ts
rchiodo: Hopefully everything will work the same as it does in LSP. The client provides the content of open files. The server reads closed files from disk. Changes to disk (to force the type server to update any internal caches) will come in the form of a notification.

docs/tsp/typeServerProtocol.ts
rchiodo: ID that uniquely identifies the current state. 

The way I have it working right now with Pyright is that any invalidation of the cache causes the internal snapshot number to increment. So a snapshot is just a number. And it really just tracks if there's been any changes detected since the snapshot was handed out. 

It's not intended to be an actual snapshot of the state. If the snapshot doesn't match, all requests should return undefined for their results or throw an exception. Haven't worked that out yet. Exception is probably better. LSP already has this idea for canceled exceptions. We could add a new error result number that indicates invalid snapshot.

I'll add more comments to indicate that the server is not required to handle more than one snapshot at a time.

docs/tsp/typeServerProtocol.ts
rchiodo: Yes. I'll change the comment.

docs/tsp/typeServerProtocol.ts
rchiodo: I had it here originally but it seemed redundant. Well because there's only ever one version, the most recent one. But maybe that's an implementation detail. 

If a type server did support multiple snapshots for some reason, you might want the version of an older snapshot. I'll add the snapshot back.

docs/tsp/typeServerProtocol.ts
```diff
@@ -0,0 +1,383 @@
...
+        return connection.onRequest(method, handler);
+    }
+
+    export function onNotification<P extends Params, M extends Notifications & keyof P & string>(
+        connection: NotificationReceiver,
+        method: M,
+        handler: NotificationHandler<P[M]>
+    ): Disposable {
+        return connection.onNotification(method, handler);
+    }
```
rchiodo, line 382: @erictraut, thanks a lot for all of your feedback. Internally I'm going to try and make a lot of the changes you suggested and see how it affects the semantic token provider.  I'll include you on the review list on the next update to the public version.

docs/tsp/typeServerProtocol.ts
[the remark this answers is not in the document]
rchiodo: Ah thanks. Good idea.

docs/tsp/typeServerProtocol.ts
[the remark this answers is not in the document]
erictraut: `BuiltIn` is used internally to pyright as a shorthand for "this class comes from a handful of stdlib typeshed stubs". I don't think it should appear in the TSP. The language server code should be modified to use the namespace of the class to distinguish it from other classes of the same name. So, rather than using `ClassType.isBuiltIn`, it should use the name + moduleName of the class to check for `builtins.str`.

docs/tsp/typeServerProtocol.ts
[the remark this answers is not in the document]
erictraut: This flag is an implementation detail of pyright, and its meaning may change over time. Currently, the type `type[T]` (where `T` is a type variable) has the category of `TypeVar` but has the `Instantiable` flag set. I've considered changing it in the future to model it as a Class `type` with a type argument. This alternative way to model it has a number of advantages and avoids a bunch of special cases in pyright's current logic. It's a big change though. My point is that the language server code currently makes a number of assumptions that are rooted in pyright's modeling of types, and those assumptions will likely not hold for other type checkers. If the goal is to create a type-checker-neutral interface, you're going to need to create an abstraction that eliminates all of these assumptions.

docs/tsp/typeServerProtocol.ts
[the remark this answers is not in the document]
erictraut: > bound to a class

That's not what the flag means in pyright currently. Pyright uses this flag to indicate that it's a method that has been decorated with the `@staticmethod` decorator and therefore cannot be bound to a class.

docs/tsp/typeServerProtocol.ts
erictraut: > only ever one version, the most recent one

That works if the interface is synchronous because you can ask for the diagnostic version immediately after asking for the diagnostics and be assured that it hasn't changed in the meantime. But the TSP interface is asynchronous, so that assumption no longer holds.

docs/tsp/typeServerProtocol.ts
erictraut: It's probably better if it is documented to return only the overloaded signatures (i.e. filter them out). This could help facilitate collapsing the `Function` and `Overloaded` categories into a single category.

docs/tsp/typeServerProtocol.ts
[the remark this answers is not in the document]
rchiodo: They are sort of bound to a transaction now. The snapshot idea. Once the snapshot increments, all type handles are no longer valid. 

And the type server need only keep handles to types it has returned, not all of its internal types.

Internally I'm just using an array that gets pushed when a new handle is created. An array of WeakReferences. The array is cleared when the snapshot has cause to update.

Do you think we need a more explicit transaction boundary? Like a start/stop request? That makes handling multiple requests at the same time more difficult though.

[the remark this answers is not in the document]
erictraut: Hmm, maybe that would work. What causes a snapshot to increment, and who controls when this occurs? Isn't that effectively dictated by the user — by when he or she makes a modification to a file that's tracked by the type server? Do you think this will be a problem if a large number of language server operations are performed without any file modifications being made?

You mentioned that you're using an array of WeakReferences. That means these references may be deallocated, which means that subsequent lookups could fail. Is that a problem, or is this just something that the language server needs to be prepared to handle?

PR #7241, opened 2025-05-19
Update type server protocol based on feedback from @erictraut

rchiodo:
Next round of updates based on feedback Eric had. Thanks for the input Eric.

For more information about the type server protocol and what it's for, see this discussion item:
https://github.com/microsoft/pylance-release/discussions/7180

/cc
@kinto0, @carljm, @erictraut

### DIFF
--- a/docs/tsp/typeServerProtocol.ts
+++ b/docs/tsp/typeServerProtocol.ts
@@ -46,7 +46,7 @@ export namespace TypeServerProtocol {
         // URI of the source file containing this node.
         uri: string;
         // Hash of the content of the source file at the time the AST was created.
-        contentHash: number;
+        fileContentHash: number;
         // The start byte position (zero-based) of the node in the source file.
         // Note this is per byte and not per character.
         start: number;
@@ -56,20 +56,20 @@ export namespace TypeServerProtocol {
 
     // Represents a category of a type, such as class, function, variable, etc.
     export const enum TypeCategory {
-        Class = 0,
-        Function = 1,
-        Overloaded = 2,
-        Module = 3,
-        Variable = 4,
-        Parameter = 5,
-        Property = 6,
-        TypeAlias = 7,
-        TypeVariable = 8,
-        Generic = 9,
-        Unknown = 10,
-        Any = 11,
-        None = 12,
-        Union = 13,
+        // Type can be anything
+        Any,
+        // Callable type
+        Function,
+        // Functions defined with @overload decorator
+        Overloaded,
+        // Class definition
+        Class,
+        // Module instance
+        Module,
+        // Union of two or more other types
+        Union,
+        // Type variable
+        TypeVar,
     }
 
     // Flags that describe the characteristics of a type.
@@ -79,10 +79,9 @@ export namespace TypeServerProtocol {
         Instantiable = 1 << 0, // Indicates if the type can be instantiated.
         Instance = 1 << 1, // Indicates if the type represents an instance (as opposed to a class or type itself).
         Callable = 1 << 2, // Indicates if an instance of the type can be called like a function. (It has a `__call__` method).
-        BuiltIn = 1 << 3, // Indicates if the type is a built-in type (like `int`, `str`, etc.).
-        Literal = 1 << 4, // Indicates if the instance is a literal (like `42`, `"hello"`, etc.).
-        Interface = 1 << 5, // Indicates if the type is an interface (a type that defines a set of methods and properties). In Python this would be a Protocol.
-        Generic = 1 << 6, // Indicates if the type is a generic type (a type that can be parameterized with other types).
+        Literal = 1 << 3, // Indicates if the instance is a literal (like `42`, `"hello"`, etc.).
+        Interface = 1 << 4, // Indicates if the type is an interface (a type that defines a set of methods and properties). In Python this would be a Protocol.
+        Generic = 1 << 5, // Indicates if the type is a generic type (a type that can be parameterized with other types).
     }
 
     // Flags that describe the characteristics of a function or method.
@@ -92,7 +91,7 @@ export namespace TypeServerProtocol {
         Async = 1 << 0, // Indicates if the function is asynchronous.
         Generator = 1 << 1, // Indicates if the function is a generator (can yield values).
         Abstract = 1 << 2, // Indicates if the function is abstract (must be implemented in a subclass).
-        Static = 1 << 3, // Indicates if the function is static (belongs to the class rather than an instance).
+        Static = 1 << 3, // Indicates if the function has a @staticmethod decorator.
     }
 
     // Flags that describe the characteristics of a class.
@@ -103,10 +102,16 @@ export namespace TypeServerProtocol {
     }
 
     export interface Type {
-        // Unique identifier for the type definition within the session.
+        // Unique identifier for the type definition within the snapshot. A handle doesn't need to exist beyond
+        // the lifetime of the snapshot.
+        //
+        // It can be used by the Type Server to reference internal information about a type when the
+        // type is returned in a subsequent request.
+        //
+        // When the internal snapshot is changed, all outstanding handles should be invalid and no longer usable.
         handle: string | number;
 
-        // Essential classification of the symbol.
+        // Essential classification of the Type.
         category: TypeCategory;
 
         // Flags describing the type.
@@ -115,43 +120,45 @@ export namespace TypeServerProtocol {
         // Name of the module the type comes from
         moduleName: string | undefined;
 
-        // Name of the type.
+        // Simple name of the type. For example, for a class `MyClass` in module `my_module`, this would be `MyClass`.
         name: string;
 
         // Flags specific to the category
         categoryFlags: number;
     }
 
-    export interface Member {
-        // The name of the member.
+    export interface Attribute {
+        // The name of the attribute.
         name: string;
 
-        // The type of the member.
+        // The type of the attribute.
         type: Type;
 
-        // The type the member came from
+        // The type the attribute came from
         classType: Type | undefined;
+
+        // The type the attribute is bound to, if applicable.
+        boundType: Type | undefined;
     }
 
-    // Flags that are used for searching for members of a type.
-    export const enum MemberAccessFlags {
+    // Flags that are used for searching for attributes of a class Type.
+    export const enum AttributeAccessFlags {
         None = 0,
-        SkipInstanceMembers = 1 << 0, // Skip instance members when searching for members of a type.
-        SkipObjectBaseClass = 1 << 1, // Skip members from the `object` base class when searching for members of a type.
+        SkipInstanceAttributes = 1 << 0, // Skip instance attributes when searching for attributes of a type.
+        GetBoundAttributes = 1 << 1, // Look for bound attributes when searching for attributes of a type. That is methods bound specifically to an instance.
     }
 
     // Represents the category of a declaration in the type system.
     // This is used to classify declarations such as variables, functions, classes, etc.
     export const enum DeclarationCategory {
-        Intrinsic = 0, // An intrinsic is a built-in type or function that is part of the language itself, such as `int`, `str`, `len`, etc.
-        Variable = 1, // A variable is a named storage location that can hold a value.
-        Param = 2, // A parameter is a variable that is passed to a function or method.
-        TypeParam = 3, // A type parameter is a placeholder for a type that can be specified when the type is used, such as in generics.
-        TypeAlias = 4, // A type alias is a name that refers to another type, such as `List[int]` or `Dict[str, int]`.
-        Function = 5, // A function is a named block of code that can be called with arguments and returns a value.
-        Class = 6, // A class is a blueprint for creating objects that encapsulates data and behavior.
-        SpecialBuiltInClass = 7, // Special built-in types like 'Tuple', 'Generic',  'Protocol', 'Callable', 'Type', etc
-        Alias = 8, // An alias declaration, which is a reference to another declaration.
+        Intrinsic, // An intrinsic refers to a symbol that has no actual declaration in the source code, such as built-in types or functions. One such example is a '__class__' declaration.
+        Variable, // A variable is a named storage location that can hold a value.
+        Param, // A parameter is a variable that is passed to a function or method.
+        TypeParam, // This is for PEP 695 type parameters.
+        TypeAlias, // This is for PEP 695 type aliases.
+        Function, // A function is any construct that begins with the `def` keyword and has a body, which can be called with arguments.
+        Class, // A class is any construct that begins with the `class` keyword and has a body, which can be instantiated.
+        Import, // An import declaration, which is a reference to another module.
     }
 
     // Represents a symbol declaration in the type system.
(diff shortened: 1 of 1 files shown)


docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
-        // Request to get the effective meta class of a type.
-        GetEffectiveMetaClass = 'typeServer/getEffectiveMetaClass',
+        // Request to get the meta class of a type.
+        GetMetaclass = 'typeServer/getMetaclass',
         // Request to get the type of a declaration.
         GetTypeOfDeclaration = 'typeServer/getTypeOfDeclaration',
         // Request to get symbol declaration information for a node.
         GetSymbolDeclarationInfo = 'typeServer/getSymbolDeclarationInfo',
-        // Request to get the bound magic method for a type.
-        GetBoundMagicMethod = 'typeServer/getBoundMagicMethod',
```
rchiodo, line 269: I got rid of this all together. getTypeAttribute works the same way if I pass in the 'bound' flag.

debonte (approved):

kinto0:
Thanks for all the hard work!! It’s amazing you were able to get this working so quickly. It looks like a very promising technology with a lot of potential. I left a few inline comments on specific details below. 
As we look into actually implementing this in a type checker, we are finding a mountain of work ahead of us. It could be more work than LSP would be. 

Stepping back, TSP has the potential to benefit our type checker in two ways: 
1. Letting us skip implementation of certain features because TSP handles it for us
2. Letting type checker authors take advantage of fancy Pylance features for free (the stuff it adds on top of Pyright)

Benefit one fails if it takes longer to implement the feature with TSP than it would with LSP. I worry about 1) time spent because a typechecker is not set up in a similar-enough way to what TSP asks for and 2) time spent implementing the actual protocol request/responses itself. 

I wonder if using a superset/wrapper of LSP is a way to partially solve both of these concerns:
1. We expect type checkers to eventually implement full LSP to support other editors, so they should be optimized for performance using the usual protocol
2. There already exist LSP server libraries in most languages

If Pylance talked to a type checker over LSP (and the checker was fully implemented), how can Pylance add value?
- It can rank autocomplete
- It can format raw documentation + provide links in docstrings
- It can use copilot with results
- … The list goes on

This doesn’t work for every feature: some require back/forth communication. I’m not sure what the answer for those is.
What about when a feature is not yet implemented in a type checker? Example: Pylance asks for call hierarchy, but a type checker has not yet implemented it. 

Let’s say the type checker has implemented find-references and go-to definition. The checker would respond with “not implemented” for call hierarchy, and Pylance would know to ask the server for repeated find-references and go-to definitions until it can build a full call-hierarchy response. 

Some other open questions:
- Initialization: it still seems TSP will still need some sort of capabilities/versioning handshake to ensure backwards compatibility
- Execution: How will these type servers be started? Will they still live in their own VSCode extension which would give the checker author control for updating it?

docs/tsp/typeServerProtocol.ts
```diff
@@ -56,20 +56,20 @@ export namespace TypeServerProtocol {
...
+        Class,
+        // Module instance
+        Module,
+        // Union of two or more other types
+        Union,
+        // Type variable
+        TypeVar,
     }
 
     // Flags that describe the characteristics of a type.
```
kinto0, line 75: cc @ndmitchell

does this make assumptions about how we represent these types that won't be compatible with pyrefly? how will be know if it's missing anything?

docs/tsp/typeServerProtocol.ts
```diff
@@ -115,43 +120,45 @@ export namespace TypeServerProtocol {
         // Name of the module the type comes from
```
kinto0, line 120: declaration's moduleName comment mentions this:
```
// The dot-separated import name for the file that
// contains the declaration (may not be definitive
// because a source file can be accessed via different
// import names in some cases).
```

I imagine moduleName means the same thing here? If it, should both of these be a `ModuleName` type which is a list of `nameParts`?

This makes assumptions about the project root. TBD if problems arise from this as I'm looking at the rest of this.

docs/tsp/typeServerProtocol.ts
```diff
@@ -115,43 +120,45 @@ export namespace TypeServerProtocol {
...
         moduleName: string | undefined;
 
-        // Name of the type.
+        // Simple name of the type. For example, for a class `MyClass` in module `my_module`, this would be `MyClass`.
         name: string;
 
         // Flags specific to the category
         categoryFlags: number;
```
kinto0, line 127: what flags are possible here?

docs/tsp/typeServerProtocol.ts
```diff
@@ -115,43 +120,45 @@ export namespace TypeServerProtocol {
...
+        Variable, // A variable is a named storage location that can hold a value.
+        Param, // A parameter is a variable that is passed to a function or method.
+        TypeParam, // This is for PEP 695 type parameters.
+        TypeAlias, // This is for PEP 695 type aliases.
+        Function, // A function is any construct that begins with the `def` keyword and has a body, which can be called with arguments.
+        Class, // A class is any construct that begins with the `class` keyword and has a body, which can be instantiated.
+        Import, // An import declaration, which is a reference to another module.
     }
 
     // Represents a symbol declaration in the type system.
```
kinto0, line 164: (comment for node?: Node line)

are the only Declarations that shouldn't have `node` `DeclarationCategory.Intrinsic`? if so, can we codify that in the type system?

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
-        Shutdown = 'typeServer/shutdown',
+        ShutDown = 'typeServer/shutdown',
         // Request from client to get the current snapshot of the type server.
         // A snapshot is a point-in-time representation of the type server's state, including all loaded files and their types.
         // A type server should change its snapshot whenever any type it might have returned is no longer valid. Meaning types are
         // only usable for the snapshot they were returned with.
+        //
+        // Snapshots are not meant to survive any changes that would make the type server throw away its internal cache. They are merely an
+        // identifier to indicate to the client that the type server will accept requests for types from that snapshot.
         GetSnapshot = 'typeServer/getSnapshot',
```
kinto0, line 280: when the request is for a snapshot that is no longer valid, I imagine we will return undefined for every request. How will the client know to ask for a new snapshot?

Is the internal state only meant to be modified from TSP messages like `typeServer/filesChanged` (and that is when the client asks for a new snapshot)?

Pyrefly's state allows for multiple clients to connect to it with multiple "unsaved changes" states but only a single truth for state on-disk (@neil confirmation requested).

Additionally, our internal build of pyrefly can use other file watchers like watchman that might support different file systems. 

I wonder if we will need a way to push snapshot updates from the type server.

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
         // Request to get diagnostics for a specific file.
         GetDiagnostics = 'typeServer/getDiagnostics',
```
kinto0, line 282: Similar question: what if the server wants to push new diagnostics to the client? We might see file watch events that VSCode's file watcher does not.

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
         // Request to get symbol declaration information for a node.
         GetSymbolDeclarationInfo = 'typeServer/getSymbolDeclarationInfo',
-        // Request to get the bound magic method for a type.
-        GetBoundMagicMethod = 'typeServer/getBoundMagicMethod',
-        // Request to resolve an alias declaration.
-        // This request is sent by the client to resolve an alias declaration, which may involve looking up the
-        // declaration in the type server's cache or performing additional resolution logic.
-        ResolveAliasDeclaration = 'typeServer/resolveAliasDeclaration',
+        // Request to resolve an import declaration. Example: `from module import something`. The `something` is the import declaration.
+        ResolveImportDeclaration = 'typeServer/resolveImportDeclaration',
```
kinto0, line 300: I don't fully understand why this is necessary

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
         // Request to resolve an import.
         ResolveImport = 'typeServer/resolveImport',
```
kinto0, line 302: I don't understand this one

docs/tsp/typeServerProtocol.ts
```diff
@@ -230,6 +244,20 @@ export namespace TypeServerProtocol {
...
 
+    export interface GetTypeAttributeParams {
+        // The type for which the attribute is being requested.
+        type: Type;
+        // The name of the attribute being requested.
+        attributeName: string;
+        // Flags that control how the attribute is accessed.
+        accessFlags: AttributeAccessFlags;
+        // Optional: The expression node that the member is being accessed from.
+        expressionNode?: Node;
```
kinto0, line 255: when is this necessary? why is it necessary?

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
         GetSnapshot = 'typeServer/getSnapshot',
         // Request to get diagnostics for a specific file.
         GetDiagnostics = 'typeServer/getDiagnostics',
         // Request to get the version of diagnostics for a specific file.
         GetDiagnosticsVersion = 'typeServer/getDiagnosticsVersion',
         // Request to get the type information for a specific node.
         GetType = 'typeServer/getType',
-        // Request to get the union types for a specific type.
+        // Request to get the collection of subtypes that make up a union type.
         GetUnionTypes = 'typeServer/getUnionTypes',
```
kinto0, line 288: should this instead be typed in a way where we are guaranteed `Type` is of category `UnionType`?

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
-        // Request to find a member of a type.
-        GetTypeMember = 'typeServer/getTypeMember',
-        // Request to get all overloads of a function or method.
+        // Request to find an attribute of a class.
+        GetTypeAttribute = 'typeServer/getTypeAttribute',
+        // Request to get all overloads of a function or method. The returned value doesn't include the implementation signature.
         GetOverloads = 'typeServer/getOverloads',
```
kinto0, line 292: @ndmitchell is it easy to get overloads for a type? does this have performance implications?

docs/tsp/typeServerProtocol.ts
```diff
@@ -241,36 +269,35 @@ export namespace TypeServerProtocol {
...
-        // Request to get the effective meta class of a type.
-        GetEffectiveMetaClass = 'typeServer/getEffectiveMetaClass',
+        // Request to get the meta class of a type.
+        GetMetaclass = 'typeServer/getMetaclass',
         // Request to get the type of a declaration.
         GetTypeOfDeclaration = 'typeServer/getTypeOfDeclaration',
```
kinto0, line 296: seems like we can avoid this API by using the `Node` from the `Declaration`. I think the only case it won't work is for `Intrinsic`.

docs/tsp/typeServerProtocol.ts
```diff
@@ -288,42 +315,33 @@ export namespace TypeServerProtocol {
         SettingsChange = 'typeServer/settingsChange',
```
kinto0, line 315: our typechecker settings are partially represented by VSCode settings (which I guess will have to be watched somehow in this protocol) and partially represented by config files. How will the type client understand when any arbitrary configuration changes? For this reason, I think it makes more sense to have a "push" based snapshot approach where the type server pushes these.

docs/tsp/typeServerProtocol.ts
rchiodo: Not sure if you're asking @ndmitchell or me this question? This set of categories seems pretty generic but if you didn't support them all, you'd just return Any for things you don't have more specific information on.

docs/tsp/typeServerProtocol.ts
rchiodo: Good idea. I'll incorporate that next time. Make this a name parts list.

docs/tsp/typeServerProtocol.ts
rchiodo: FunctionFlags if a function and ClassFlags if a class. I should make that more explicit in the comment here.

docs/tsp/typeServerProtocol.ts
rchiodo: I don't think json is descriptive enough to do such a thing. If all of these things are typescript, then we could do type unions but since these types have to be represented more generically, I don't think we can do anything other than commenting when it's undefined.

docs/tsp/typeServerProtocol.ts
rchiodo: It gives a clue to the type checker on how to look up the attribute. Specifically, when looking up bound members. 

For example, this class heirarchy:

```python
class A:
    def foo(self):
        pass
        
class B(A):
     def foo(self):
         pass
         
b: A = B()
b.foo() # Get type attributes for foo here
```

docs/tsp/typeServerProtocol.ts
rchiodo: Oh that sounds like a good idea. A notification that suggests snapshot changed. I'll add that to the notification list.

docs/tsp/typeServerProtocol.ts
rchiodo: We could handle this with another notification, diagnostics changed or something along those lines.

docs/tsp/typeServerProtocol.ts
rchiodo: I don't think we can do inheritance with json types which is why everything is so flat.

docs/tsp/typeServerProtocol.ts
rchiodo: The declaration in this case points to the symbol 'something'. That symbol isn't defined in the current module. This function goes and finds the symbol in the other module.

Here's the location of this function in the pyright source:

https://github.com/microsoft/pyright/blob/5021af8f0545386b99e916b764bb8968d362bd7f/packages/pyright-internal/src/analyzer/declarationUtils.ts#L231

it's basically finding the original symbol.

Maybe the name of this request should be findOriginalDecl or something like that instead?

docs/tsp/typeServerProtocol.ts
rchiodo: ResolveImport? That is used for finding the source file for where the type server thinks an import is being pulled from. 

Example:

```
import os
```

ResolveImport on 'os' would return the on disk path to where the os.py file is.

docs/tsp/typeServerProtocol.ts
rchiodo: Yeah this is a little backwards at the moment. The pylance extension tells VS code which configurations it is watching. But then it asks for the settings instead of being sent them. This should probably be changed to same pattern.

The Type server should query for settings, it shouldn't be told them.

rchiodo:
For this question:

>  it still seems TSP will still need some sort of capabilities/versioning handshake to ensure backwards compatibility

I agree. Just haven't added it yet. The initial version (implemented by Pyright) there are no unsupported capabilities, but we can certainly fine grain this more. Maybe per request even.

For this question:

> How will these type servers be started? Will they still live in their own VSCode extension which would give the checker author control for updating it?

I think we can figure this out later but I was imagining the extension of the type server would have to be installed and the user would select it somehow. We could then either embed the knowledge of how to start them in Pylance or create some extension api  that the other extension implements that describes how to start them.

docs/tsp/typeServerProtocol.ts
rchiodo: Pyright's [implementation](https://github.com/microsoft/pyright/blob/5021af8f0545386b99e916b764bb8968d362bd7f/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L22197) of this is more involved than the same thing for a node. It could certainly be implemented internally using the node but a declaration seems like it can get more information.

docs/tsp/typeServerProtocol.ts
kinto0: for us, something closer to LSP like resolveDefinition would be better since it's already implemented: can we ask for the definition of imports themselves?

docs/tsp/typeServerProtocol.ts
kinto0: in this case, we can figure this out with just type: `B` and attributeName: `foo`. what extra information does expression node provide?

docs/tsp/typeServerProtocol.ts
rchiodo: Not sure what you mean by the definition of the import? Do you mean goto def?
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition

That's one of the use cases for this message internally. We also use it to determine if a module is 3rd party code or not.

rchiodo: We also use it on the language server side to build up source file trees which determines which files we search through for stuff like find all references.

docs/tsp/typeServerProtocol.ts
rchiodo: Maybe that was a bad example, this is more of the case that hits:

```python
class A:
    def __bool__(self) -> bool:
        return True
        
a = bool(A()) # Goto def on bool here
print(a)
```

Internally we search for the `__bool__` method. It needs the A() in the expression to know what to search for.

rchiodo: The `bool(A())` is the expression node in this case.

PR #7358, opened 2025-06-25
Changelog, etc for 2025.6.2

bschnurr:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 2025.6.2 (25 June 2025) Release
```
rchiodo, line 3: @debonte is this the right order? By date?

debonte: I think in the past we've sorted by version number because that's what people are generally searching for.

Issue #7523, opened 2025-08-29
Why not listen on 127.0.0.1 instead of randomly opening a port for 0.0.0.0, which poses a security risk


kelegeek:
Why not listen on 127.0.0.1 instead of randomly opening a port for 0.0.0.0, which poses a security risk

rchiodo:
@heejaechang it looks like this is the MCP server.

I can see this for the server.bundle.js process (process id 54532 for me) when I run netstat:

```
C:\Users\rchiodo>netstat -aon | findstr 54532
  TCP    0.0.0.0:3566           0.0.0.0:0              LISTENING       54532
  TCP    127.0.0.1:3566         127.0.0.1:59415        ESTABLISHED     54532
  TCP    [::]:3566              [::]:0                 LISTENING       54532
  UDP    0.0.0.0:54532          *:*                                    51844
```

And in our log:

```
2025-08-29 09:14:50.266 [info] (54532) Starting Mcp server id:3566 ...
```

Is the cors express by default listening on 0.0.0.0 even though we only allow localhost?

```
        const app = express();

        // Middleware to parse JSON bodies
        app.use(express.json());

        // Only accept connections from localhost
        app.use(cors({ origin: ['http://localhost'] }));
```

heejaechang:
let me take a look.

rchiodo:
This issue has been fixed in version 2025.8.2, which we've just released. You can find the changelog here: [CHANGELOG.md](https://github.com/microsoft/pylance-release/blob/main/CHANGELOG.md#202582-18-September-2025-Release)

PR #7562, opened 2025-09-17
Changelog, etc for 2025.8.1

bschnurr:


rchiodo (approved):
There's actually a bug we fixed that we didn't mention. This one:
https://github.com/microsoft/pylance-release/issues/7523

PR #7975, opened 2026-04-08
TROUBLESHOOTING.md: mention meson-python incompatibility with compat_…

klimanskiy1:
This PR updates TROUBLESHOOTING.md to include a note about meson-python and its limitations with editable installs. Users relying on meson-python may encounter unresolved import warnings in Pylance because this build system does not support compat_mode=strict or compat_mode=compat, unlike pip/setuptools.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -69,6 +69,12 @@ default. It will only use import hooks if you set `dev-mode-exact` to `true`.
 files by default. It will only use import hooks if you set `editable-backend` to
 `"editables"`.
 
+#### meson-python
+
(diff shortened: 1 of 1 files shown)


heejaechang:
they could manually resolve it with extra path setting?

klimanskiy1:
Technically yes, via `python.analysis.extraPaths`, but that’s just a workaround and not fully reliable. It won’t behave the same as a proper editable install and can still cause type checking inconsistencies.

heejaechang:
I would still at least mention about extra path as a workaround?

klimanskiy1:
Got it

klimanskiy1:
Done!

rchiodo (approved):

rchiodo:
Thanks for the PR

Issue #5364, opened 2024-01-17
Globs in extra paths


judej:
### Discussed in https://github.com/microsoft/pylance-release/discussions/2712

<div type='discussions-op-text'>

<sup>Originally posted by **itsdani** February 21, 2021</sup>
I'm trying to setup a python monorepo build with bazel, and I'm struggling to make pylance recognize the library imports. I have managed to make it work by adding e.g. `<bazel-myproject>/external/<my-pip-dependencies>/pypi__numpy` to `python.analysis.extraPaths`, but adding a line for each dependency on every computer we use doesn't seem to be right.

My idea/request is that if we could add something like `dependencies/*/` or `dependencies/pypi__*/` to the extra paths, it would solve this problem entirely. The expected behavior would be to add every (matching) directory under `dependencies/` to the extra paths.</div>

luabud:
we just converted this back to an issue to add it to our roadmap

dpar39:
Any ETA on this?

erictraut:
As has been explained in other threads, I don't think it will work to add globs for extraPaths. The problem is that extraPaths is order dependent. The order affects import resolution behavior. Globs is an order-independent mechanism. If you want to add extra paths to your project, you should specify them one at a time in the order you intend for imports to be resolved.

rchiodo:
We could make it deterministic by saying globs resolve alphabetically.

heejaechang:
@itsdani @dpar39 it has been a while but can you guys provide us more detail on the scenario? and we have `add extra path` code action to help these kinds of situation and wonder why that didn't work?

itsdani:
We have given up on using Bazel and I work on very different things since I requested it 3 years ago, I'm not even using Pylance anymore. Maybe someone currently experiencing this issue could provide more detail. I can only repeat what I already said in the other thread: https://github.com/microsoft/pylance-release/discussions/2712#discussioncomment-6799676

adzenith:
Bazel puts each Python dependency in its own directory for sandboxing reasons. Our extraPaths is 601 lines long, and we have to manually update it when our Python dependency tree changes.

Here's a small snippet so you can see what it looks like:
```
  "external/rules_python~~pip~pip_310_google_re2/site-packages",
  "external/rules_python~~pip~pip_310_google_resumable_media/site-packages",
  "external/rules_python~~pip~pip_310_googleapis_common_protos/site-packages",
  "external/rules_python~~pip~pip_310_graphviz/site-packages",
[...]

PR #8111, opened 2026-07-08
docs: add how-to for glob patterns in extraPaths

heejaechang:
Adds `docs/howto/extra-paths-glob-resolution.md`, a user guide for glob pattern support in `python.analysis.extraPaths` — covering the wildcard syntax (`*`, `**`, `?`), deterministic ordering and precedence, configuration across VS Code settings and config files, performance/memory considerations, and troubleshooting.

Also updates related docs to reflect that globs are now supported:

- `docs/settings/python_analysis_extraPaths.md`: updates the wildcards/globs FAQ from "not supported" to supported, with the determinism and performance rationale.
- `docs/howto/monorepo-setup.md`, `docs/howto/reading-pylance-logs.md`, `docs/howto/performance-tuning.md`: behavioral notes and cross-links.
- `docs/INDEX.md`: registers the new guide.

Related issue: https://github.com/microsoft/pylance-release/issues/5364

### DIFF
--- a/docs/howto/extra-paths-glob-resolution.md
+++ b/docs/howto/extra-paths-glob-resolution.md
@@ -61,11 +61,24 @@ Because `extraPaths` is order-sensitive (the first matching root wins during imp
 
 Two rules govern the result:
 
-1. **Sort within a glob.** The directories a single glob matches are inserted at that glob's position, sorted in **ascending order** by their resolved path. The comparison is case-sensitive and **ordinal** (it compares Unicode code points and normalizes to NFC), so it never depends on your locale, language, or operating system — Windows, macOS, and Linux produce an identical order.
+1. **Sort within a glob.** The directories a single glob matches are inserted at that glob's position, sorted in **ascending order** by their resolved path. The comparison is **ordinal**: paths are compared one Unicode code point at a time (after normalizing to NFC), never with a language-aware collation. Two consequences are worth calling out:
+    - **"Case-sensitive" means no case-folding.** Uppercase `A`–`Z` (U+0041–U+005A) all sort before lowercase `a`–`z` (U+0061–U+007A), so `Auth` comes before `api`. Names that differ only in case are kept as separate directories, because case can change the module name they map to.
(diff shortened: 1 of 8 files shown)


docs/howto/extra-paths-glob-resolution.md
```diff
@@ -0,0 +1,351 @@
...
+{
+    "python.analysis.extraPaths": [
+        "stubs",
+        "packages/core/src",
+        "packages/*/src",
+        "vendor/proto/python",
+        "vendor/*/python"
+    ]
+}
+```
```
rchiodo, line 148: I feel like an example showing case sensitive ordering would be nice to add. It's the one I'm not sure how it will work. 

And what does case sensitive mean with unicode characters? Who's case sensitive algorithm is picked? Like for japanese kanji, there is 'reading' based sorting and "stroke" based sorting and other  different sorting algorithms. Not sure what case sensitive even means in those cases.

rchiodo (approved):

docs/howto/extra-paths-glob-resolution.md
heejaechang: Good question — I expanded the ordering rule and added an example to cover exactly this ([f64a2a3](https://github.com/microsoft/pylance-release/commit/f64a2a3)).

Short version: the sort is **not** a linguistic collation, so none of the culture-specific algorithms (kanji reading vs. stroke, phonetic, dictionary, locale) are involved. Directories are compared **ordinally, one Unicode code point at a time**, after normalizing to NFC.

So:

- **"Case-sensitive" just means no case-folding.** `A`–`Z` (U+0041–U+005A) sort before `a`–`z` (U+0061–U+007A), so `Auth` comes before `api`. Names that differ only in case stay distinct (case can change the module name they map to).
- **For non-ASCII names** (accented Latin, CJK, emoji, …) the order is by Unicode code-point value. It's fully deterministic and identical on every machine/locale/OS, but it is intentionally *not* "alphabetical" in any human-language sense — there's no reading/stroke/phonetic ordering to pick between.

The doc now spells this out and includes a worked example showing `Core`, `Zeta` sorting ahead of `api`, `auth`.

arnavkumarkvs-prog:
sir can you guide me i am making a asissant and there is tons of erorr 
you guidence matter a lot

Issue #8230, opened 2026-09-21
Auto import suggestions (still) don't work when *.pth file is present


pdc1:
I am sorry to repost this issue, but I believe my previous issue #8225 was closed prematurely.

It was closed based on this comment: `Pth files are only supported if we can download a CPython WASI build that matches your python build. Otherwise, we can't run the code in the pth files`

This should not be required. Based on the link provided in that response (https://github.com/microsoft/pylance-release/blob/main/docs/howto/editable-installs.md) my case is:

Build Backend | Default .pth Type | Pylance Works? | How to Get Path-Based .pth
--------------------- | ------------------------- | ---------------------- | ----------------------------------------
Poetry | Path-based | Yes | Default behavior

Per this issue? Pylance works? => **No**

It sounds like the WASI module is only needed if "python.analysis.enableEditableInstalls" = true, which is not the case for my environment.

Please re-open this issue and help me understand why **path-based** PTH files are interfering with auto import suggestion.

pdc1:
In case it is not clear from the linked issue, the PTH file in my case does not contain any code, it is a literal path to the project: `/tank/projects/pth-auto-import`.

rchiodo:
I can't seem to reproduce the problem. This is using poetry. 

<img width="778" height="348" alt="Image" src="https://github.com/user-attachments/assets/fda254db-ad89-487b-93e5-89793b4ce3c6" />

Imports do show up for the installed modules. My pth file has a static path:

```
# /home/rchiodo/.cache/pypoetry/virtualenvs/portfolio-gIlBC-eH-py3.13/lib/python3.13/site-packages/portfolio.pth
/home/rchiodo/pylance-repros/issue-8230-4486a556-poetry
```

Maybe my setup is different. Here's the modules I created:

<img width="404" height="463" alt="Image" src="https://github.com/user-attachments/assets/f7eed42b-5e01-43bd-b0e8-300f4d74a48b" />

rchiodo:
Oh it's this setting right here: 

```
    "python.analysis.showOnlyDirectDependenciesInAutoImport": true
```

Modules from the pth file are considered 3rd party imports so it skips the files from it.

rchiodo:
With that setting false:

<img width="635" height="274" alt="Image" src="https://github.com/user-attachments/assets/e9034b93-5484-41af-aeaa-f2c15538ebe1" />

With that setting true:

[...]

PR #8231, opened 2026-09-21
Document editable-install auto-import filtering and full-mode defaults

rchiodo:
## Summary

- Explain that `python.analysis.showOnlyDirectDependenciesInAutoImport` can omit auto-import completions for path-based `.pth` editable installs, including Poetry installs, even when their source is in the workspace.
- Clarify that `python.analysis.languageServerMode: "full"` enables this filter by default, and show how to retain full mode while explicitly disabling the filter.
- Add cross-linked troubleshooting guidance distinguishing completion filtering from import resolution and the "Add import" code action, while noting that indexing and other completion settings still apply.

Related to #8230. Documentation-only clarification; no runtime behavior changes.

## Validation

- `git diff --check`
- Verified relative links and anchors across the three changed documents.
- Parsed JSON examples and checked the full-mode example's explicit `false` override.


pdc1:
As the author of the linked issue, these edits look great! Thanks for the additional explanation and documentation.

heejaechang (approved):

